PR #363, opened 2024-04-22
ceremonial merge of merges

bghira:


bghira:
there's no reason to implement anything special for 1.5... just use 'legacy'.

elismasilva:
> there's no reason to implement anything special for 1.5... just use 'legacy'.

For that, yes, what I referred to was the training script. I created a train_sd15.py with some modifications to dreambooth and for deepspeed. But i'm not finished yet. So for: "--model_type"  which I referred to could be called "--training_type"

parser.add_argument(
        "--model_type",
        type=str,        
        choices=[
            "full",
            "lora",
            "deepfloyd-full",
            "deepfloyd-lora",
            "deepfloyd-stage2",
            "deepfloyd-stage2-lora",
            "dreambooth"  <---
        ],
        default="full",
        help=(
            "The training type to use. 'full' will train the full model, while 'lora' will train the LoRA model."
            " LoRA is a smaller model that can be used for faster training."
        ),
    )

But that's ok, it's just another experiment for me, you said you're not interested in 1.5, I'm not going to push these changes.

elismasilva:
I understand, I haven't had the opportunity to test the DoRa and Lycoris styles yet, but have you tested any training in 1.5 with this 2.0 script? I ran some training steps and didn't see any changes in the model like the dreambooth, perhaps it requires more extensive training, I need to test these other new styles. For now, due to my hardware limitations, the only one that worked well with quality was the Dreambooth in WSL and deepspeed. But like I said, I'm doing a laboratory test, if it's successful I'll share the results later. Regarding code maintenance, I agree with you, having too many files gets in the way. On the other hand, unifying everything in a single script can become a headache in the future as a change can affect the behavior of other things that you didn't want to have changed. To unify, you might have to think about segmenting the code with classes, using inheritance, something like that, but also only if you intend to grow and evolve the scripts. I thought about making a GUI interface with gradio for it would also help for those who are used to Kohya and one trainer.

bghira:
for SD 1.5, it actually is very slow to train. i'm not sure if you're basing it on a SD 1.5 finetune or if you're continuing the CompVis / runwayML weights.

- generally, community finetune weights are in fp16 precision which are unsuitable for continuing training
  - they may involve offset or pyramidal noise, both of which are problematic
- the base SD 1.5 model is going to need hundreds of thousands of steps to train, as it takes forever to learn
  - compare to SD 2.x which learns concepts very quickly even without text encoder finetuning

there's many many settings that can influence whether the model learns or not;

- kohya disables max grad clipping (scaling) by default, but simpletuner has `--max_grad_norm=2`
- using EMA will make it nearly impossible to pick up changes without big compute behind the training
- a low learning rate, especially with a high batch size (eg. gradient accumulation steps * train_batch_size)

a UI:
- for now, tools that build upon this with a UI are likely to break frequently because there's no stable SDK for them to interact with. things change a lot and i have no way to communicate that to any UI toolkit.
  - not to say this isn't desirable, but until the trainer's base functionality is stabilised (eg. post v1.0) i think the workload increase of fielding issue reports from users that are legitimately requiring a UI to use the trainer with, will be too much.
  - i like how bmaltais and kohya's projects are separate, such that the UI issues go to an entirely different person. but **this is a major problem** for the long term sustainability of the project(s) as bmaltais i think ends up having a "stale" version of kohya for UI users.
  - there's also derrian distro which seems to introduce even more problems for kohya users, as it modifies defaults or hides functionality.

elismasilva:
> for SD 1.5, it actually is very slow to train. i'm not sure if you're basing it on a SD 1.5 finetune or if you're continuing the CompVis / runwayML weights.
> 
> * generally, community finetune weights are in fp16 precision which are unsuitable for continuing training
>   
>   * they may involve offset or pyramidal noise, both of which are problematic
> * the base SD 1.5 model is going to need hundreds of thousands of steps to train, as it takes forever to learn
>   
>   * compare to SD 2.x which learns concepts very quickly even without text encoder finetuning
> 
> there's many many settings that can influence whether the model learns or not;
> 
> * kohya disables max grad clipping (scaling) by default, but simpletuner has `--max_grad_norm=2`
> * using EMA will make it nearly impossible to pick up changes without big compute behind the training
> * a low learning rate, especially with a high batch size (eg. gradient accumulation steps * train_batch_size)
> 
> a UI:
> 
> * for now, tools that build upon this with a UI are likely to break frequently because there's no stable SDK for them to interact with. things change a lot and i have no way to communicate that to any UI toolkit.
>   
>   * not to say this isn't desirable, but until the trainer's base functionality is stabilised (eg. post v1.0) i think the workload increase of fielding issue reports from users that are legitimately requiring a UI to use the trainer with, will be too much.
>   * i like how bmaltais and kohya's projects are separate, such that the UI issues go to an entirely different person. but **this is a major problem** for the long term sustainability of the project(s) as bmaltais i think ends up having a "stale" version of kohya for UI users.
>   * there's also derrian distro which seems to introduce even more problems for kohya users, as it modifies defaults or hides functionality.

Most of the problems I see regarding SDKs are that people are constantly updating project dependencies. I am of the opinion that if you have found a configuration that is working you should lock the dependencies, you can only update a dependency if there are serious security risks or if the update brought some benefit related to quality and performance. Any new dependency update requires you to regression test everything and it is not just about breaking code or not but also about the expected results which is more difficult and tedious to perform.

bghira:
i think updates should be performed regularly. if results start to deviate wildly it's because the update came, rather than months down the line trying to reconcile multiple major version upgrades at once.

for example, apple MPS builds for simpletuner rely on pytorch nightly 2.4. i'd have everything on nightly builds if NVIDIA weren't such a pain.

elismasilva:
It's difficult to balance your plates when you have multiplatform environments. There are situations where it sometimes makes sense to have a separate version just for a specific platform, otherwise you will have to keep stitching the code and running into irreconcilable problems.

bghira:
i come from the OpenZFS world, which is a kernel driver for a filesystem that has many complex features - using `ifdef` for multi-platform support just feels natural :P

PR #373, opened 2024-05-01
AMD/ROCm support

Beinsezii:
Barebones, based on the Mac config.

Verified with
```sh
pip install -U poetry
poetry install --no-root -C install/rocm
bash train_sdxl.sh
```

As it's based on the `apple` config, some libs aren't included ATM that possibly work on ROCm. Let me know if there's other important libs I can trial.

Added DeepSpeed 0.14 as it compiles fine on my GPU w/ ROCm 6

CC @bghira


Beinsezii:
Also torch is pinned up to 2.3 because lower versions will fail the dependency resolve on ROCm 6, and ROCm versions < 6.0 have page fault issues on my GPU.

bghira:
maybe a fused backwards pass would help rocm and mps use less memory

bghira (approved):
sounds good to me

Beinsezii:
What is `triton-library` used for? That's the only library I couldn't get working on ROCm besides `xformers`.

I greped the whole repo and didn't see it imported anywhere.

bghira:
it's used by torch itself to compile kernels

Beinsezii:
`triton-library`, not `triton`? `triton` is included in the pytorch ROCm repository already.

bghira:
thank you

PR #446, opened 2024-06-13
sd3 timestep selection weighting changes

bghira:


bghira:
differs from upstream diffusers in that i have tested the previous mode weighting and found it was correct

PR #451, opened 2024-06-14
Unload LoRA weights before deleting pipeline object

tolgacangoz:
Hi @bghira, great library!
I thought this would be proper.


bghira:
oof, good catch. thank you!

tolgacangoz:
Thanks for merging!

PR #466, opened 2024-06-16
Updated documentation formatting as well as added on to the QUICKSTART.md

TwoAbove:
Hey! 

Fixed the formatting of the .md files a bit, and also iterated on the `QUICKSTART.md`. 

Let me know if I missed anything!


documentation/DATALOADER.md
```diff
@@ -140,15 +156,15 @@ A complete example list can be found [here](/caption_filter_list.example.txt). I
...
+arafed
 this .* has a
 ^this is the beginning of the string
 s/this/will be found and replaced/
 ```
 
 In order, the lines behave as follows:
 
-- `arafed ` (with a space at the end) will be removed from any caption it is found in. Including a space at the end means the caption will look nicer, as double-spaces won't remain. This is unnecessary, but it looks nice.
+- `arafed` (with a space at the end) will be removed from any caption it is found in. Including a space at the end means the caption will look nicer, as double-spaces won't remain. This is unnecessary, but it looks nice.
```
bghira: could you return the space here? it's important as part of the demonstration.

documentation/DATALOADER.md
```diff
@@ -256,43 +268,141 @@ This is the most common configuration, and list of aspect buckets trainable for
...
+  "2.6": [1664, 640],
+  "4.38": [2240, 512],
+  "0.42": [704, 1664],
+  "0.89": [1024, 1152],
+  "2.7": [1728, 640],
+  "5.0": [2240, 448],
+  "0.44": [704, 1600],
+  "0.94": [1024, 1088],
+  "2.8": [1792, 640],
+  "5.14": [2304, 448]
```
bghira: this might be way too long? i had it column-wise to save vertical scrolling.

documentation/QUICKSTART.md
```diff
@@ -2,135 +2,116 @@
...
+```bash
+cp sdxl-env.sh.example sdxl-env.sh
+```
+
+There, you will need to modify the following variables:
+
+- `MODEL_TYPE` - Set this to `lora`.
+- `STABLE_DIFFUSION_3` - Set this to `true`.
+- `MODEL_NAME` - Set this to `stabilityai/stable-diffusion-3-medium-diffusers`. Note that you will need to log in to Huggingface and be granted access to download this model. We will go over logging in to Huggingface later in this tutorial.
+- `BASE_DIR` - Set this to the directory where you want to store your outputs and datasets. It's recommended to use a full path here.
```
bghira, line 65: there were a couple other important keys like bitfit=false, dora=false

documentation/QUICKSTART.md
```diff
@@ -2,135 +2,116 @@
...
+
+There are a few more if using a Mac M-series machine:
+
+- `MIXED_PRECISION` should be set to `no`.
+- `USE_XFORMERS` should be set to `false`.
+
+#### Dataset considerations
+
+It's crucial to have a substantial dataset to train your model on. There are limitations on the dataset size, and you will need to ensure that your dataset is large enough to train your model effectively. Note that the bare minimum dataset size is `TRAIN_BATCH_SIZE * GRADIENT_ACCUMULATION_STEPS` as well as more than `VAE_BATCH_SIZE`. The dataset will not be useable if it is too small.
```
bghira, line 74: VAE batch size is an "up to" value. it will encode fewer items

documentation/QUICKSTART.md
```diff
@@ -2,135 +2,116 @@
...
+
+Depending on the dataset you have, you will need to set up your dataset directory and dataloader configuration file differently. In this example, we will be using two datasets: the `pseudo-camera-10k` dataset and a local dataset.
```
bghira: it's just one backend really, but the text embeds go to their own location for efficiency, as the same prompt might exist in multiple datasets, it will share a single cache object

documentation/QUICKSTART.md
```diff
@@ -139,34 +120,30 @@ mkdir -p datasets
 pushd datasets
     git clone https://huggingface.co/datasets/ptx0/pseudo-camera-10k
 popd
-```
+````
```
bghira: four backticks, needs to be 3

bghira (changes requested):
thank you. just a few things to clean up

documentation/QUICKSTART.md
TwoAbove: So this should be worded as `as well as divisible by VAE_BATCH_SIZE`?

documentation/QUICKSTART.md
TwoAbove: In `sdxl-env.sh.example` `USE_DORA` is already `false` and `USE_BITFIT` is automatically set to `false`

documentation/QUICKSTART.md
bghira: no it doesn't have to be divisible at all - you can have vae batch size at 100 and a single image dataset, and that one image will encode by itself.

documentation/QUICKSTART.md
bghira: it actually has a conditional statement in there and it enables bitfit by default for full model training.

but i guess your guide is for lora, so this is ok

documentation/QUICKSTART.md
```diff
@@ -2,135 +2,116 @@
...
+There are a few more if using a Mac M-series machine:
+
+- `MIXED_PRECISION` should be set to `no`.
+- `USE_XFORMERS` should be set to `false`.
+
+#### Dataset considerations
+
+It's crucial to have a substantial dataset to train your model on. There are limitations on the dataset size, and you will need to ensure that your dataset is large enough to train your model effectively. Note that the bare minimum dataset size is `TRAIN_BATCH_SIZE * GRADIENT_ACCUMULATION_STEPS` as well as more than `VAE_BATCH_SIZE`. The dataset will not be useable if it is too small.
+
+Depending on the dataset you have, you will need to set up your dataset directory and dataloader configuration file differently. In this example, we will be using `pseudo-camera-10k` as the dataset.
```
bghira: actually let's change this `block-out` code segment to a [link] to the dataset https://huggingface.co/datasets/ptx0/pseudo-camera-10k

documentation/QUICKSTART.md
```diff
@@ -2,135 +2,116 @@
...
 ]
 ```
 
-#### Download dataset
+The, navigate to the `BASE_DIR` directory and create a `datasets` directory:
 
+```bash
 Execute the following commands:
 
 ```bash
```
bghira: i think this might begin the block twice?

bghira: image

bghira: Then, *

as well

bghira (approved):
just a few final fixes needed and then lgtm

documentation/QUICKSTART.md
TwoAbove: Ahh, how did I miss that, thanks!

bghira:
danks!

PR #468, opened 2024-06-16
sd3 QoL improvements | better error messages | documentation improvements, quickstart guide | vae cache: prevent corruption on multigpu systems | apple: mixed precision auto override for convenience | sd3: remove text encoder training from LoRA, as it doesn't work with MM-DiT

bghira:


dill-shower:
Sorry for the necropost. But have there been any changes since then? Why aren't we training text encoders for SD3/3.5? Clip L, g, T5xxl. Don't text encoder models need to be trained for new concepts?

bghira:
nope, text encoder training requires hinge loss (eg. real CLIP training) and putting image gradients through T5 tends to break it. there's not much value to tuning the text encoder the way most image trainers do it. you require contrastive image pairs to push the CLIP model to and from a single concept and its opposites. that is what the "contrastive" part is.

PR #469, opened 2024-06-16
Add SD3 model extension code

AmericanPresidentJimmyCarter:
- Adds the ability to use a larger SD3 with qk norm added back (as LayerNorm)
- Executing the module allows you to convert an existing SD3 module to the expanded module

### DIFF
--- a/helpers/sd3/expanded.py
+++ b/helpers/sd3/expanded.py
@@ -1,0 +1,630 @@
+import argparse
+import gc
+import operator
+import os
+
+from typing import Any, Dict, Optional, Union
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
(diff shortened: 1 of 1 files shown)


bghira:
🫡

FullBleed:
Is there a discussion somewhere about this?

AmericanPresidentJimmyCarter:
> Is there a discussion somewhere about this?

You can discuss it here. The model is expanded according to the [SOLAR model methodoloy](https://arxiv.org/html/2312.15166v2). After finetuning for about 1000 steps the model looks more or less normal, but to fully take advantage of the expanded model you would probably need to continue finetuning for a while on a very large amount of data.

Given the current licensing problems with SD3, there doesn't seem to be much of an appetite to do this, despite it appearing to work.

PR #471, opened 2024-06-17
pixart sigma tuning

bghira:
- [x] text encoder cache
- [x] vae cache
- [x] training loop model predictions
- [x] basic argument parsing
- [x] saving the model checkpoints
- [x] loading model checkpoints
- [x] validation inference time
- [x] documentation


TwoAbove:
Hey one thing I noticed is that a lot of things now use `train_sdxl.*` scripts. It may make sense to rename it to something like `train_model.*`.

Issue #481, opened 2024-06-18
(question)SD3 Training Memory Usage Issue


pzzmyc:
Regarding the memory usage issue during SD3 training, it starts at 9GB and goes up to 24GB. I tried adding code to clear the cache, which alleviated the issue, but I'm not sure if it's correct. Here is the code I added to clear the cache every 100 steps: 
                ...
                current_epoch_step += 1
                StateTracker.set_global_step(global_step)
                **if global_step % 100 == 0:
                    torch.cuda.empty_cache()
                    gc.collect()
                    print("clear cache")**

                ema_decay_value = "None (EMA not in use)"
                if args.use_ema:
                ....

bghira:
does it OOM? if not, and it's in the cache, it's evictable, which means it's free-able upon need

sometimes the kernel might block a little longer than usual which a low-latency application might decide isn't good enough and give an OOM error instead of waiting.

pzzmyc:
Yes, I am using 24GB graphics cards. Sometimes I encounter a CUDA out of memory error, and sometimes an NCCL error, which is likely also caused by insufficient GPU memory

bghira:
would you like to open a PR for this to optionally run based on an interval?

pzzmyc:
> would you like to open a PR for this to optionally run based on an interval?

I would like to, but I'm not very familiar with these operations on GitHub. How do I do it?

bghira:
- fork the project to your github account
- pull a local copy of it to make changes from
- create a new branch, and commit your changes
- push them to your repository
- open the pull request to this repository

bghira:
please set `--accelerator_cache_clear_interval=500` in your TRAINER_EXTRA_ARGS to enable the new option from @TwoAbove

PR #489, opened 2024-06-19
Added CUDA GC every x steps

TwoAbove:
This is a preliminary PR for [#481](https://github.com/bghira/SimpleTuner/issues/481)

With the `garbage_collection` function, I checked a lot of repos, and this looks like the most common sequence.

I want to dedupe the `garbage_collection` function from the `train_*` scripts, but I don't have the intuition to put it into the correct spot. Can you help here @bghira? 

Thanks!

### DIFF
--- a/helpers/arguments.py
+++ b/helpers/arguments.py
@@ -1428,11 +1428,11 @@ def parse_args(input_args=None):
         ),
     )
     parser.add_argument(
-        "--cuda_clear_cache_steps",
+        "--accelerator_cache_clear_interval",
         default=None,
         type=int,
         help=(
-            "Clear the CUDA cache every X steps. This can help prevent memory leaks, but may slow down training."
+            "Clear the cache from VRAM every X steps. This can help prevent memory leaks, but may slow down training."
         ),
     )
 
--- a/helpers/arguments.py
+++ b/helpers/arguments.py
@@ -1427,6 +1427,14 @@ def parse_args(input_args=None):
             " This is a safety feature to prevent accidental use of an unsupported optimizer, as weights are stored in bfloat16."
         ),
     )
+    parser.add_argument(
+        "--cuda_clear_cache_steps",
+        default=None,
+        type=int,
+        help=(
+            "Clear the CUDA cache every X steps. This can help prevent memory leaks, but may slow down training."
+        ),
+    )
 
     if input_args is not None:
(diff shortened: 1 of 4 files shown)


helpers/arguments.py
```diff
@@ -1427,6 +1427,14 @@ def parse_args(input_args=None):
             " This is a safety feature to prevent accidental use of an unsupported optimizer, as weights are stored in bfloat16."
         ),
     )
+    parser.add_argument(
+        "--cuda_clear_cache_steps",
```
bghira: we may want it called accelerator_cache_clear_interval

train_sd21.py
```diff
@@ -109,6 +109,15 @@
 check_min_version("0.27.0.dev0")
 
 
+def garbage_collection():
+    import gc
+
+    if torch.cuda.is_available():
+        torch.cuda.empty_cache()
+        torch.cuda.ipc_collect()
+    gc.collect()
```
bghira: we can also call the torch.mps equivalents

this should probably be in `helpers/caching/memory.py` (just create it)

it can be called `reclaim_memory()` to match kohya's code.

TwoAbove: Nice!

helpers/caching/memory.py
```diff
@@ -0,0 +1,12 @@
+def reclaim_memory():
+    import gc
```
TwoAbove, line 2: Will using imports like this have a performance penalty? 
I guess even if it does, the garbage collection will take orders of magnitude more time.

bghira: we import torch in threads everywhere, i'm pretty sure it's fine :D

helpers/caching/memory.py
```diff
@@ -0,0 +1,12 @@
...
+    import torch
+
+    if torch.cuda.is_available():
+        torch.cuda.empty_cache()
+        torch.cuda.ipc_collect()
+
+    if torch.backends.mps.is_available():
+        torch.mps.empty_cache()
```
bghira, line 10: i *think* it needs mps.synchronize() too

bghira: https://pytorch.org/docs/stable/generated/torch.mps.synchronize.html

bghira (approved):

bghira:
ty!

PR #493, opened 2024-06-19
Fixed GC

TwoAbove:
Missed this arg when renaming it


TwoAbove:
Oh you fixed it, cool

PR #524, opened 2024-06-22
Hunyuan-DiT feature branch plan

bghira:
We're going to add Hunyuan DiT to its own feature branch, and this issue encompasses the plan to integrate this model.

If you wish to take one of the components, simply comment here and open a pull request to this branch (`feature/hunyuan-dit`). Please do not open your pull request to the main branch.

- [x] a `--hunyuan_dit` argment is added
  - this occurs in multiple places. look for `pixart_sigma` to get an idea of where these occur.
  - leave things like text embed cache broken, until they are properly implemented - so that a ValueError etc is raised when something still needs to be "plugged in".
  - we have to then `StateTracker.set_model_type("hunyuan_dit")`
- [x] model_type gets updated to include hunyuan_dit value alongside sdxl, sd3, pixart_sigma, etc
  - this applies when `--hunyuan_dit` argument is provided
  - a `HUNYUAN_DIT=false` variable export should be added to `sdxl-env.sh.example`
  - `train_sdxl.sh` should be updated to address this variable by setting `--hunyuan_dit` when HUNYUAN_DIT=true. see `PIXART_SIGMA` in there for info.
- [x] initial model loading for the trained transformer must be added or included alongside eg. pixart's loading, if the two are compatible and can be shared
  - this occurs at the top of `train_sdxl.py` where we load `PixArtTransformer2DModel` etc
- [ ] text embed caching then needs a hunyuan codepath added, which can probably be shared with pixart depending on how it expects the embeds to be used. for example, DeepFloyd's T5 embeds don't return the attn mask, they just encode the embeds with it. but PixArt Sigma requires the attn mask as an input to the transformer itself.
  - i don't blame anyone if they just leave this for me to implement. `helpers/caching/sdxl_embeds.py` must be augmented to include an `if-else` codepath for `hunyuan_dit` text embedding
- [x] checkpoint save/load should be enhanced to include the logic for hunyuan or share with pixart where appropriate
  - these occur inside `helpers/sdxl/save_hooks.py`, and LoRA vs Full models are handled differently
- [ ] any kind of specific loss augmentation - scaling, weighting, masking, must be added to the training loop if required
- [x] validations must load the Hunyuan DiT pipeline classes (`helpers/training/validation.py`)
  - more than likely this will require wholesale copy-pasting the Hunyuan pipeline from Diffusers into `helpers/hunyuan_dit/pipeline.py` so that we can apply dtype/device assignment corrections. see `helpers/sdxl/pipeline.py` and `helpers/sd3/pipeline.py` for more information.
- [ ] documentation must be produced and included, eg. a quickstart guide w/ a functional dataset is required for all new models added
- [ ] update model card metadata to include Hunyuan-DiT details for code example
- [x] block access to controlnet training
- [x] block access to LoRA training
- [x] block access to text encoder training
- [ ] block access to `--validation_using_datasets` if Hunyuan-DiT does not have img2img pipeline

## nice-to-have, but not needed

- [ ] controlnet training / inference validations
- [ ] LoRA training
- [ ] refiner validations, eg. `denoising_start` support in pipeline

## testing

- [ ] loading initial model
- [ ] saving a checkpoint
- [ ] loading a checkpoint
- [ ] final pipeline export
- [ ] text2img validations
- [ ] img2img validations
- [ ] if implemented, refiner-based partial denoising validations (uses PIL inputs, sorry)
- [ ] i think the vae cache works just fine as-is, but it should be tested
- [ ] min-snr-gamma compatibility
- [ ] whether offset noise should be allowed
- [ ] noise scheduler changes at validation time via `--validation_noise_scheduler`


sayakpaul:
I will try to tackle the first two in a PR next week early and get a PR ready for you.

MohamedAliRashad:
@bghira I think seperating different models to different files will be nicer.
I see the `sdxl` files are being used for training SD3 and now Hunyuan and this can be confusing to new commers

bghira:
it will be renamed eventually, but it's not that big of a deal imo.

sayakpaul:
Will tackle "initial model loading for the trained transformer must be added or included alongside eg. pixart's loading, if the two are compatible and can be shared" next.

sayakpaul:
Sure. Thanks for being willing. 

Could you provide some entry points for me that tackle model loading? That will be a bit easier and I will get cracking.

bghira:
it will be in `train_sdxl.py` at the top where we make a new PixArtTransformer2DModel

this is separate from the later task to add checkpoint save/load. this just sets the models up from their base counterparts

bghira:
i've updated this branch with latest commits from main.

sayakpaul:
@bghira in 05ae7ac I swapped the hooks to move the LoRA refactors out. Additionally, I have created https://github.com/bghira/SimpleTuner/pull/554 to allow other PRs to build on top of this refactor. Makes sense?

bghira:
@sayakpaul i've rebased this to the main branch.

sayakpaul:
Thanks a bunch. The failing tests seem to be unrelated.

bghira:
yes, just some network issue on the runner

bghira:
update and rebased this PR ontop of the Kolors support @sayakpaul

bghira:
@sayakpaul still interested in hunyuan-dit?

sayakpaul:
I am. Particularly because of the multi-lingual component to it. Will pick it up soon.

bghira:
@sayakpaul can't get the actual model prediction to work. but i've done the rest of the needful to make validations happen. they don't look normal on MacOS. it's all just noise. so i can't verify them here locally

bghira:
@sayakpaul i will assume there is no longer interest in this but feel free to pick it up again at some point if you like

PR #527, opened 2024-06-24
multi-caption

bghira:


helpers/image_manipulation/training_sample.py
```diff
@@ -691,7 +691,7 @@ def get_conditioning_image(self):
         conditioning_dataset = StateTracker.get_conditioning_dataset(
             data_backend_id=self.data_backend_id
         )
-        print(f"Conditioning dataset components: {conditioning_dataset.keys()}")
+        # print(f"Conditioning dataset components: {conditioning_dataset.keys()}")
```
sayakpaul, line 694: Maybe we can just log this information formally?

helpers/metadata/backends/parquet.py
```diff
@@ -136,6 +136,12 @@ def _extract_captions_to_fast_list(self):
                 filename = str(row[filename_column])
             else:
                 filename = str(index)
+            if type(caption_column) == list:
+                caption = None
+                if len(caption_column) > 0:
+                    caption = [row[c] for c in caption_column]
+            else:
+                caption = row[caption_column]
```
sayakpaul, line 147: Do we wanna randomize the caption sampling part similar to [this](https://github.com/huggingface/diffusers/blob/4ad7a1f5fd48ce3aabc46f090f28b56a02e733aa/examples/text_to_image/train_text_to_image.py#L780)?

helpers/prompts.py
```diff
@@ -255,7 +254,10 @@ def prepare_instance_prompt_from_parquet(
         if image_caption:
             image_caption = image_caption.strip()
         if prepend_instance_prompt:
-            image_caption = instance_prompt + " " + image_caption
+            if type(image_caption) == list:
+                image_caption = [instance_prompt + " " + x for x in image_caption]
+            else:
+                image_caption = instance_prompt + " " + image_caption
```
sayakpaul, line 260: Could be nice to have a proper template class that could accept style and subject keywords along with the main body of the caption. So, for example:

image

(Figure from https://arxiv.org/abs/2402.12004)

sayakpaul (commented):
From the looks of it, I don't see anything obviously weird. What are we supporting here to be specific?

helpers/prompts.py
bghira: i'm not sure if { } can be in a filename, can it?

helpers/metadata/backends/parquet.py
bghira: this job is to retrieve all captions for caching

helpers/image_manipulation/training_sample.py
bghira: it wasn't needed, but it's nice to have sometimes commented out like this if i'm adding a new model or training / validation style

bghira:
this is to support datasets with multiple captions per sample via `textfile` caption strategy which loads them from `.txt` files. and the parquet backend. but i'm waiting for it to actually be tested with a multi-caption dataset.

PR #529, opened 2024-06-24
Fix missing SD3 import

Disty0:
SDXLSaveHook uses `SD3Transformer2DModel` at init (line 106) but `SD3Transformer2DModel` isn't imported before init and the code fails.
This PR removes `SD3Transformer2DModel` import from load_model_hook and moves it to the main imports alongside SD3 Pipeline.


bghira:
i'd already fixed this in a devel branch, but forgot to push. thanks for your PR!

PR #538, opened 2024-06-27
Support Hunyuan DiT loading

sayakpaul:
I took the liberty to refactor the LoRA related bits in the hooks but LMK if you would like me to discard that. 

Additionally, it might be nice to configure some formatting and linting on the repository. Do you have recommendations? I use this: https://gist.github.com/sayakpaul/a48a0f63d0e346164a97c687f27c5b51.

### DIFF
--- a/helpers/sdxl/save_hooks.py
+++ b/helpers/sdxl/save_hooks.py
@@ -116,7 +116,7 @@ class SDXLSaveHook:
             elif self.args.hunyuan_dit:
                 self.ema_model_cls = HunyuanDiT2DModel
 
-    def handle_lora(self, models, weights, output_dir):
+    def save_lora(self, models, weights, output_dir):
         # for SDXL/others, there are only two options here. Either are just the unet attn processor layers
         # or there are the unet and text encoder atten layers.
         unet_lora_layers_to_save = None
@@ -191,7 +191,7 @@ class SDXLSaveHook:
             os.path.join(output_dir, "training_state.json")
         )
         if "lora" in self.args.model_type:
-            self.handle_lora(models=model, weights=weights, output_dir=output_dir)
+            self.save_lora(models=model, weights=weights, output_dir=output_dir)
             return
 
         # Create a temporary directory for atomic saves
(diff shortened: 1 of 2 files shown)


helpers/sdxl/save_hooks.py
```diff
@@ -106,79 +113,85 @@ def __init__(
...
+            ):
+                text_encoder_3_lora_layers_to_save = (
+                    convert_state_dict_to_diffusers(
+                        get_peft_model_state_dict(model)
                     )
+                )
 
-                elif isinstance(
+            elif not isinstance(model, type(unwrap_model(self.accelerator, HunyuanDiT2DModel))):
+                if isinstance(
```
bghira, line 159: we go from `if isinstance()` to `elif not isinstance()` ? it looks like we never save the hunyuan-dit weights

sayakpaul: Because the codepath below is for LoRA and Hunyuan DiT doesn’t support LoRA yet. Or am I reading it wrong?

bghira:
git pre-commit hook:

```sh
#!/bin/bash

export CHANGED_FILES
CHANGED_FILES=$(git diff --cached --name-only --diff-filter=ACM | grep -e '\.py$')
# Format each changed file:
for file in $CHANGED_FILES; do
    black "$file" || exit 1
    git add "$file"
done

python -m unittest discover -s tests/ || exit 1

exit 0
```

helpers/sdxl/save_hooks.py
bghira: since you'd refactored it, i assumed there was support :P

sayakpaul: Hunyuan DiT PEFT support is first needed in diffusers. Once done, will add here.

sayakpaul:
So, it’s just black at the moment. We can move to ruff since it’s faster. Can make a PR for it if you would like?

bghira:
i haven't had any performance issues with black, but i'm not cleaning up 100k files every time either :D

sayakpaul:
Fair. That is understandable.

PR #553, opened 2024-07-02
implement Aura Diffusion training

bghira:
This issue encompasses the plan to support Aura Diffusion.

If you wish to take one of the components, simply comment here and open a pull request to this branch (`feature/aura-diffusion`). Please do not open your pull request to the main branch.

- [x] a `--aura_diffusion` argment is added
  - this occurs in multiple places. look for `pixart_sigma` to get an idea of where these occur.
  - leave things like text embed cache broken, until they are properly implemented - so that a ValueError etc is raised when something still needs to be "plugged in".
  - we have to then `StateTracker.set_model_type("aura_diffusion")`
- [x] model_type gets updated to include aura_diffusion value alongside sdxl, sd3, pixart_sigma, etc
  - this applies when `--aura_diffusion` argument is provided
  - a `AURA_DIFFUSION=false` variable export should be added to `sdxl-env.sh.example`
  - `train_sdxl.sh` should be updated to address this variable by setting `--aura_diffusion` when AURA_DIFFUSION=true. see `PIXART_SIGMA` in there for info.
- [x] initial model loading for the trained transformer must be added or included alongside eg. pixart's loading, if the two are compatible and can be shared
  - this occurs at the top of `train_sdxl.py` where we load `PixArtTransformer2DModel` etc
- [x] text embed caching then needs n aura_diffusion codepath added, which can probably be shared with pixart depending on how it expects the embeds to be used. for example, DeepFloyd's T5 embeds don't return the attn mask, they just encode the embeds with it. but PixArt Sigma requires the attn mask as an input to the transformer itself.
- [x] checkpoint save/load should be enhanced to include the logic for aura_diffusion or share with pixart where appropriate
  - these occur inside `helpers/sdxl/save_hooks.py`, and LoRA vs Full models are handled differently
- [x] any kind of specific loss augmentation - scaling, weighting, masking, must be added to the training loop if required
- [x] validations must load the Aura Diffusion pipeline classes (`helpers/training/validation.py`)
  - more than likely this will require wholesale copy-pasting the Aura pipeline from Diffusers into `helpers/aura_diffusion/pipeline.py` so that we can apply dtype/device assignment corrections. see `helpers/sdxl/pipeline.py` and `helpers/sd3/pipeline.py` for more information.
- [x] documentation must be produced and included, eg. a quickstart guide w/ a functional dataset is required for all new models added
- [x] update model card metadata to include Aura Diffusion details for code example
- [x] block access to controlnet training
- [x] block access to LoRA training
- [x] block access to text encoder training
- [x] block access to `--validation_using_datasets` if Aura Diffusion does not have img2img pipeline

## nice-to-have, but not needed

- [ ] controlnet training / inference validations
- [ ] LoRA training
- [ ] refiner validations, eg. `denoising_start` support in pipeline

## testing

- [x] loading initial model
- [x] saving a checkpoint
- [x] loading a checkpoint
- [ ] final pipeline export
- [x] text2img validations
- [ ] img2img validations
- [ ] if implemented, refiner-based partial denoising validations (uses PIL inputs, sorry)
- [x] i think the vae cache works just fine as-is, but it should be tested
- [ ] ~~min-snr-gamma compatibility~~
- [ ] ~~whether offset noise should be allowed~~
- [ ] ~~noise scheduler changes at validation time via `--validation_noise_scheduler`~~

cc @sayakpaul should we pull your save/load lora refactor into a separate PR so that this can build on it?

### DIFF
--- a/OPTIONS.md
+++ b/OPTIONS.md
@@ -27,6 +27,11 @@ This guide provides a user-friendly breakdown of the command-line options availa
 - **What**: Enable PixArt Sigma training quirks/overrides.
 - **Why**: PixArt is similar to SD3 and DeepFloyd in one way or another, and needs special treatment at validation, training, and inference time. Use this option to enable PixArt training support. PixArt does not support ControlNet, LoRA, or `--validation_using_datasets`
 
+### `--aura_flow`
+
+- **What**: Enable AuraFlow training quirks/overrides.
+- **Why**: As a flow-matching model, AuraFlow has several unique needs. This option must be enabled to load and train an AuraFlow model.
+
 **Note:** Like SDXL and SD3, PixArt Sigma **also** uses the `train_sdxl.sh`/`train_sdxl.py` training script, `sdxl-env.sh` configuration file.
 
 ### `--pretrained_model_name_or_path`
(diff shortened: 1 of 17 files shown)


sayakpaul:
> cc @sayakpaul should we pull your save/load lora refactor into a separate PR so that this can build on it?

@bghira sure, we cut that part out. I don't mind.

bghira:
though most of the items are ticked it is just because their stubs are in place. there are `NotImplementedError`s peppered throughout the codebase that must be replaced with their corrected implementations.

bghira:
well, it's mostly done. i'll test it tomorrow.

Issue #546, opened 2024-06-28
error validating


woweenie:
Training SD3 lora.

"[ERROR] (helpers.training.validation) Error generating validation image: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu! (when checking argument for argument mat1 in method wrapper_CUDA_addmm), Traceback (most recent call last):
  File "/workspace/train-sd3/SimpleTuner/helpers/training/validation.py", line 762, in validate_prompt
    validation_image_results = self.pipeline(**pipeline_kwargs).images"

If I put the VALIDATION_STEPS down to 10 then it works, but at the deafult 150 it just prints the error

woweenie:
i think it happens when the VALIDATION_STEPS is longer than the number of steps int he epoch...

OliviaOliveiira:
Had the very same error, and simple reboot fixed it for me

mhirki:
I'm also seeing this issue but it's random. If I restart the script, the error may not show up at all. If I pass the option --validation_seed_source=gpu, I get a different error that is still random.

I ran the script with SIMPLETUNER_LOG_LEVEL=DEBUG and grabbed a debug.log where the error shows up. I also grabbed another debug.log where the error was not showing up. I didn't really see anything different between these myself though.

[debug-validation-error.zip](https://github.com/user-attachments/files/16045582/debug-validation-error.zip)
[debug-validation-success.zip](https://github.com/user-attachments/files/16045586/debug-validation-success.zip)

bghira:
it does seem to be caused by having the generator on CPU:

```
2024-06-30 20:25:23,350 [DEBUG] (helpers.training.validation) Device for generator: cpu
2024-06-30 20:25:23,350 [DEBUG] (helpers.training.validation) Device for pooled_prompt_embeds: cuda:0
2024-06-30 20:25:23,350 [DEBUG] (helpers.training.validation) Device for negative_pooled_prompt_embeds: cuda:0
2024-06-30 20:25:23,350 [DEBUG] (helpers.training.validation) Device for prompt_embeds: cuda:0
2024-06-30 20:25:23,350 [DEBUG] (helpers.training.validation) Device for negative_prompt_embeds: cuda:0
2024-06-30 20:25:23,350 [DEBUG] (helpers.training.validation) Device for vae: cuda:0
2024-06-30 20:25:23,350 [DEBUG] (helpers.training.validation) Device for text_encoder_2: cpu
[...]

PR #562, opened 2024-07-06
sd3: Fix random errors when generating validation images. (#546)

mhirki:
I managed to fix the random errors in #546 for myself at least. Debug prints revealed that C<timestep> was randomly either on the CPU or on the GPU. However, after moving C<timesteps> to the GPU, another error happened because C<latents> was also on the CPU. So then I moved C<latents> also to the GPU and the errors were gone.


bghira:
thank you

PR #571, opened 2024-07-09
Nuke the text encoders by sending them to "meta"

Disty0:
PyTorch is really stubborn and doesn't actually nuke the text encoders even if we delete them and set them to none.
Sending them to the "meta" device makes sure that they are nuked.

### DIFF
--- a/train.py
+++ b/train.py
@@ -791,6 +791,9 @@ def main():
         memory_before_unload = torch.cuda.memory_allocated() / 1024**3
         if accelerator.is_main_process:
             logger.info(f"Unloading text encoders, as they are not being trained.")
+        text_encoder_1 = text_encoder_1.to("meta") if text_encoder_1 is not None else None
+        text_encoder_2 = text_encoder_2.to("meta") if text_encoder_2 is not None else None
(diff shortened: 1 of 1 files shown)


Disty0:
> it's weird, because the vram use reports that they were removed. do you see higher reduction with this?

0 GB vs 12 GB

bghira:
maybe it's specific to SD3? here's what i see on 8x H100:

```
2024-07-09 03:17:26,842 [INFO] (__main__) After nuking text encoders from orbit, we freed 8.87 GB of VRAM. The real memories were the friends we trained a model on along the way.
```

bghira:
also for what it's worth, on Apple Silicon the amount seems oddly randomised:

```
2024-07-09 07:18:31,572 [INFO] (__main__) After nuking text encoders from orbit, we freed 2.28 GB of VRAM. The real memories were the friends we trained a model on along the way.
```

vs 

```
2024-07-09 07:14:59,260 [INFO] (__main__) After nuking text encoders from orbit, we freed 15.19 GB of VRAM. The real memories were the friends we trained a model on along the way.
```

not clear why sometimes it clears more or less.

PR #588, opened 2024-07-20
support sd3 multi prompt

syx0310:
Just like the [stable diffusion 3 ComfyUI example](https://huggingface.co/stabilityai/stable-diffusion-3-medium/blob/main/comfy_example_workflows/sd3_medium_example_workflow_multi_prompt.json) Stability AI provided.


bghira:
i dont really want to mess with sd3 text embeds because it doesnt even implement attention masking correctly. once that is fixed let me know

PR #600, opened 2024-07-26
backend for csv based datasets

bghira:


bghira:
@williamhzuk did you ever manage to get this working? it looks like it has some big issues. i have managed to sort those out, i believe.

bghira:
@williamzhuk did you ever manage to get this working? it looks like it has some big issues. i have managed to sort those out, i believe.

bghira:
a big issue encountered exists on multi-gpu systems. when you're calling `save_state()` on the csv backend, it will do this across multiple processes. each process will have its own different copy of the CSV. it seems like this functionality cannot work for this backend on multi-gpu systems.

PR #607, opened 2024-07-29
config file versioning to allow updating defaults without breaking backwards compat

bghira:


helpers/caching/text_embeds.py
```diff
@@ -1020,11 +1020,6 @@ def compute_embeddings_for_sd3_prompts(
 
         return prompt_embeds_all, add_text_embeds_all
 
-    def split_cache_between_processes(self, prompts: list):
-        # Use the accelerator to split the data
-        with self.accelerator.split_between_processes(prompts) as split_files:
-            self.prompts = split_files
```
bghira, line 1026: dead code

helpers/caching/vae.py
```diff
@@ -416,23 +415,6 @@ def _reduce_bucket(
         # )
         return relevant_files
 
-    def split_cache_between_processes(self):
-        self.local_unprocessed_files = self.discover_unprocessed_files(self.cache_dir)
```
bghira, line 420: now being assigned in discover_new_files

PR #609, opened 2024-07-30
experiment: small DiT model

bghira:


bghira:
cc @sayakpaul

helpers/caching/text_embeds.py
```diff
@@ -840,6 +842,7 @@ def compute_embeddings_for_legacy_prompts(
                         if "deepfloyd" not in StateTracker.get_args().model_type:
                             # we have to store the attn mask with the embed for pixart.
                             # aura doesn't require it, but it's just easier to keep.
+                            # smoldit requires the attn mask at inference time 💪🏽
```
sayakpaul, line 845: ❣️

sayakpaul (approved):
Thanks!

Would it have been a better idea to just depend on https://github.com/sayakpaul/smoldit? This way, it's easier to pull in the changes. No?

bghira:
yeah eventually i'd like to move the dependency out of tree but for now it's quickest to do this so i can ship it around without another moving piece

PR #611, opened 2024-08-02
Fix crash when using jsonl files

swkang-rp:
Thanks for supporting jsonl, btw jsonl file crashes here.

Followed similar pattern in `parquet.py`.


bghira:
thanks. we might actually want to use `orjsonl` eventually because it's so much faster, and supports compressed volumes.

PR #622, opened 2024-08-04
quanto: exploring low-precision training.

bghira:
it does not yet work with saving or loading the loras. or is it just an MPS issue, who knows.

also you can't resume training a lora and enable quanto on the base model. it will complain.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@ For memory-constrained systems, see the [DeepSpeed document](/documentation/DEEP
 - Most models are trainable on a 24G GPU, or even down to 16G at lower base resolutions.
   - LoRA training for PixArt, SDXL, SD3, and SD 2.x that uses less than 16G VRAM
 - DeepSpeed integration allowing for [training SDXL's full u-net on 12G of VRAM](/documentation/DEEPSPEED.md), albeit very slowly.
+- Quantised LoRA training, using low-precision base model or text encoder weights to reduce VRAM consumption while still allowing DreamBooth.
 - Optional EMA (Exponential moving average) weight network to counteract model overfitting and improve training stability. **Note:** This does not apply to LoRA.
 - Train directly from an S3-compatible storage provider, eliminating the requirement for expensive local storage. (Tested with Cloudflare R2 and Wasabi S3)
 - For only SDXL and SD 1.x/2.x, full [ControlNet model training](/documentation/CONTROLNET.md) (not ControlLoRA or ControlLite)
@@ -63,6 +64,7 @@ Preliminary training support for Flux.1 is included:
 - LoRA or full tuning via DeepSpeed ZeRO
 - ControlNet training is not yet supported
 - Train either Schnell or Dev models
+- Quantise the base model using `--base_model_precision` to `int8-quanto` or `fp8-quanto` for major memory savings
 
 See [hardware requirements](#flux1-dev-schnell).
 
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -313,3 +313,32 @@ See [DREAMBOOTH](/documentation/DREAMBOOTH.md) for a breakdown on how Dreambooth
 ### Mixture-of-Experts split-schedule model training
 
 See [MIXTURE-OF-EXPERTS](/documentation/MIXTURE_OF_EXPERTS.md) for information on how to split training over two models, such that one is responsible for composition and large details, and the other is responsible for finalising and filling in the fine details.
+
+### Quantised model training
+
+Tested on Apple and NVIDIA systems, Hugging Face Optimum-Quanto can be used to reduce the precision and VRAM requirements, training even Flux.1 on just 20GB or less.
+
+Inside your SimpleTuner venv:
+
+```bash
+pip install optimum-quanto
+```
+
+```bash
+# choices: int8-quanto, int4-quanto, int2-quanto, fp8-quanto
+# int8-quanto was tested with a single subject dreambooth LoRA.
+# fp8-quanto does not work on Apple systems. you must use int levels.
+# int2-quanto is pretty extreme and gets the whole rank-1 LoRA down to about 13.9GB VRAM.
+# may the gods have mercy on your soul, should you push things Too Far.
+export TRAINER_EXTRA_ARGS="--base_model_precision=int8-quanto"
+
+# Maybe you want the text encoders to remain full precision so your text embeds are cake.
+# We unload the text encoders before training, so, that's not an issue during training time - only during pre-caching.
+# Alternatively, you can go ham on quantisation here and run them in int4 or int8 mode, because no one can stop you.
+export TRAINER_EXTRA_ARGS="${TRAINER_EXTRA_ARGS} --text_encoder_1_precision=no_change --text_encoder_2_precision=no_change --text_encoder_3_precision=no_change"
+
+# When you're quantising the model, we're not in pure bf16 anymore.
+# Since adamw_bf16 will never work with this setup, select another optimiser.
+# I know the spelling is different than everywhere else, but we're in too deep to fix it now.
+export OPTIMIZER="adafactor" # or maybe prodigy
+```
--- a/config/config.env.example
+++ b/config/config.env.example
@@ -179,7 +179,8 @@ export USE_GRADIENT_CHECKPOINTING=true
 export ALLOW_TF32=true
 # AdamW 8Bit is a robust and lightweight choice. Adafactor might reduce memory consumption, and Dadaptation is slow and experimental.
 # AdamW is the default optimizer, but it uses a lot of memory and is slower than AdamW8Bit or Adafactor.
-# Choices: adamw, adamw8bit, adafactor, dadaptation
+# When training a quantised base model, you can't use adamw_bf16. Instead, try adafactor or adamw.
+# Choices: adamw, adamw8bit, adafactor, dadaptation, adamw_bf16
 export OPTIMIZER="adamw_bf16"
 
 
--- a/documentation/DREAMBOOTH.md
+++ b/documentation/DREAMBOOTH.md
@@ -58,6 +58,35 @@ VALIDATION_PROMPT="a photograph of subjectname"
 DATALOADER_CONFIG="multidatabackend-dreambooth.json"
 ```
 
+### Quantised model training
+
+Tested on Apple and NVIDIA systems, Hugging Face Optimum-Quanto can be used to reduce the precision and VRAM requirements.
+
+Inside your SimpleTuner venv:
+
+```bash
+pip install optimum-quanto
+```
+
+```bash
+# choices: int8-quanto, int4-quanto, int2-quanto, fp8-quanto
+# int8-quanto was tested with a single subject dreambooth LoRA.
+# fp8-quanto does not work on Apple systems. you must use int levels.
+# int2-quanto is pretty extreme and gets the whole rank-1 LoRA down to about 13.9GB VRAM.
+# may the gods have mercy on your soul, should you push things Too Far.
+export TRAINER_EXTRA_ARGS="--base_model_precision=int8-quanto"
+
+# Maybe you want the text encoders to remain full precision so your text embeds are cake.
+# We unload the text encoders before training, so, that's not an issue during training time - only during pre-caching.
+# Alternatively, you can go ham on quantisation here and run them in int4 or int8 mode, because no one can stop you.
+export TRAINER_EXTRA_ARGS="${TRAINER_EXTRA_ARGS} --text_encoder_1_precision=no_change --text_encoder_2_precision=no_change"
+
+# When you're quantising the model, we're not in pure bf16 anymore.
+# Since adamw_bf16 will never work with this setup, select another optimiser.
+# I know the spelling is different than everywhere else, but we're in too deep to fix it now.
+export OPTIMIZER="adafactor" # or maybe prodigy
+```
+
 Inside our dataloader config `multidatabackend-dreambooth.json`, it will look something like this:
 
 ```json
--- a/documentation/quickstart/FLUX.md
+++ b/documentation/quickstart/FLUX.md
@@ -84,6 +84,35 @@ There, you will need to modify the following variables:
 - `VALIDATION_GUIDANCE` - Use whatever you are used to selecting at inference time for Flux.
 - `TRAINER_EXTRA_ARGS` - Here, you can place `--lora_rank=4` if you wish to substantially reduce the size of the LoRA being trained. This can help with VRAM use.
 
+#### Quantised model training
+
+Tested on Apple and NVIDIA systems, Hugging Face Optimum-Quanto can be used to reduce the precision and VRAM requirements, training Flux on just 20GB.
+
+Inside your SimpleTuner venv:
+
+```bash
+pip install optimum-quanto
+```
+
+```bash
+# choices: int8-quanto, int4-quanto, int2-quanto, fp8-quanto
+# int8-quanto was tested with a single subject dreambooth LoRA.
+# fp8-quanto does not work on Apple systems. you must use int levels.
+# int2-quanto is pretty extreme and gets the whole rank-1 LoRA down to about 13.9GB VRAM.
+# may the gods have mercy on your soul, should you push things Too Far.
+export TRAINER_EXTRA_ARGS="--base_model_precision=int8-quanto"
+
+# Maybe you want the text encoders to remain full precision so your text embeds are cake.
+# We unload the text encoders before training, so, that's not an issue during training time - only during pre-caching.
+# Alternatively, you can go ham on quantisation here and run them in int4 or int8 mode, because no one can stop you.
+export TRAINER_EXTRA_ARGS="${TRAINER_EXTRA_ARGS} --text_encoder_1_precision=no_change --text_encoder_2_precision=no_change"
+
+# When you're quantising the model, we're not in pure bf16 anymore.
+# Since adamw_bf16 will never work with this setup, select another optimiser.
+# I know the spelling is different than everywhere else, but we're in too deep to fix it now.
+export OPTIMIZER="adafactor" # or maybe prodigy
+```
+
 
 #### Dataset considerations
 
--- a/documentation/quickstart/KOLORS.md
+++ b/documentation/quickstart/KOLORS.md
@@ -71,6 +71,35 @@ There are a few more if using a Mac M-series machine:
 - `MIXED_PRECISION` should be set to `no`.
 - `USE_XFORMERS` should be set to `false`.
 
+#### Quantised model training
+
+Tested on Apple and NVIDIA systems, Hugging Face Optimum-Quanto can be used to reduce the precision and VRAM requirements of especially ChatGLM 6B (the text encoder).
+
+Inside your SimpleTuner venv:
+
+```bash
+pip install optimum-quanto
+```
+
+```bash
+# choices: int8-quanto, int4-quanto, int2-quanto, fp8-quanto
+# int8-quanto was tested with a single subject dreambooth LoRA.
+# fp8-quanto does not work on Apple systems. you must use int levels.
+# int2-quanto is pretty extreme and gets the whole rank-1 LoRA down to about 13.9GB VRAM.
+# may the gods have mercy on your soul, should you push things Too Far.
+export TRAINER_EXTRA_ARGS="--base_model_precision=int8-quanto"
+
+# Maybe you want the text encoders to remain full precision so your text embeds are cake.
+# We unload the text encoders before training, so, that's not an issue during training time - only during pre-caching.
+# Alternatively, you can go ham on quantisation here and run them in int4 or int8 mode, because no one can stop you.
+export TRAINER_EXTRA_ARGS="${TRAINER_EXTRA_ARGS} --text_encoder_1_precision=no_change"
+
+# When you're quantising the model, we're not in pure bf16 anymore.
+# Since adamw_bf16 will never work with this setup, select another optimiser.
+# I know the spelling is different than everywhere else, but we're in too deep to fix it now.
+export OPTIMIZER="adafactor" # or maybe prodigy
+```
+
 #### Dataset considerations
 
 It's crucial to have a substantial dataset to train your model on. There are limitations on the dataset size, and you will need to ensure that your dataset is large enough to train your model effectively. Note that the bare minimum dataset size is `TRAIN_BATCH_SIZE * GRADIENT_ACCUMULATION_STEPS`. The dataset will not be discoverable by the trainer if it is too small.
--- a/documentation/quickstart/SD3.md
+++ b/documentation/quickstart/SD3.md
@@ -66,6 +66,35 @@ There are a few more if using a Mac M-series machine:
 - `MIXED_PRECISION` should be set to `no`.
 - `USE_XFORMERS` should be set to `false`.
 
+#### Quantised model training
+
+Tested on Apple and NVIDIA systems, Hugging Face Optimum-Quanto can be used to reduce the precision and VRAM requirements well below the requirements of base SDXL training.
+
+Inside your SimpleTuner venv:
+
+```bash
+pip install optimum-quanto
+```
+
+```bash
+# choices: int8-quanto, int4-quanto, int2-quanto, fp8-quanto
+# int8-quanto was tested with a single subject dreambooth LoRA.
+# fp8-quanto does not work on Apple systems. you must use int levels.
+# int2-quanto is pretty extreme and gets the whole rank-1 LoRA down to about 13.9GB VRAM.
+# may the gods have mercy on your soul, should you push things Too Far.
+export TRAINER_EXTRA_ARGS="--base_model_precision=int8-quanto"
+
+# Maybe you want the text encoders to remain full precision so your text embeds are cake.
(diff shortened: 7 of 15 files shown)


bghira:
cc @sayakpaul @kaibioinfo 

doesn't work yet at least on MPS. haven't rolled it to any nvidia machines yet

bghira:
well i got it kinda working, but adamw_bf16 is bf16-only, and explodes when it sees fp8 / int8 weights.

i got it working for training, but i have to use int2 on MPS. it goes down to 13.9GB memory for training, total.

mhirki:
I got this running on my RTX 3090. Installing optimum-quanto pulled in torch-2.4.0 so I installed newer torchvision to match. Deepspeed was also throwing errors which I solved by updating it as well.

```
pip install optimum-quanto
pip install -U torchvision deepspeed
```

Like you said, adamw_bf16 optimized wasn't working so I switched to prodigy optimizer just for fun. Training is running at about 6.2s/it. VRAM usage was initially around 23.4 GB. After adding a quick hack to get the text encoders nuked for real, VRAM usage was only about 18.8 GB. This is with the default LoRA rank of 16.

sayakpaul:
This is with FP8? How are the results?

sayakpaul:
> it does not yet work with saving or loading the loras

If you could confirm this is generally the case, that would be helpful.

sayakpaul:
> also you can't resume training a lora and enable quanto on the base model. it will complain.

Feel free to create a new thread maybe and post the trace? For convenience, we could try to restrict the exploration perhaps on a CUDA device?

mhirki:
> This is with FP8? How are the results?

Yes, it's with FP8, i.e. `--base_model_precision=fp8-quanto`.

At step 300, I'm seeing something happening in the validation image. It's gonna take a few more hours to reach MAX_NUM_STEPS of 2000.

I had some issues with validation failing due to not all tensors being on the same device. I think I managed to get it working. I'm gonna post a pull request for that later.

bghira:
great, i look forward to this fix

bghira:
i've updated this code substantially so that the text encoders (all 3 in the case of SD3) can be set separately. i'm not sure if this is practical or not. it's not required to do - by default, the text encoders will be quantised to the same level as the base model. you can disable text encoder quantisation entirely by manually applying each as `no_change`.

bghira:
> > This is with FP8? How are the results?
> 
> Yes, it's with FP8, i.e. `--base_model_precision=fp8-quanto`.
> 
> At step 300, I'm seeing something happening in the validation image. It's gonna take a few more hours to reach MAX_NUM_STEPS of 2000.
> 
> I had some issues with validation failing due to not all tensors being on the same device. I think I managed to get it working. I'm gonna post a pull request for that later.

you might want to verify you're not using int8. but maybe fp8 will be faster for you. i'm not sure - but the code was basically ignoring the request for fp8 before, and used int8 instead.

bghira:
> > also you can't resume training a lora and enable quanto on the base model. it will complain.
> 
> Feel free to create a new thread maybe and post the trace? For convenience, we could try to restrict the exploration perhaps on a CUDA device?

i narrowed it down to just float8 not being supported, Mac users can still benefit from quanto's int levels :)

sayakpaul:
A minor comment: usually int4 kernels tend to only work with bf16. Just a note to keep in mind.

mhirki:
> you might want to verify you're not using int8. but maybe fp8 will be faster for you. i'm not sure - but the code was basically ignoring the request for fp8 before, and used int8 instead.

Ok, I had a look at the code I was running and it does indeed appear to be int8 quantization in reality. I'll try fp8 later once this run finishes.

So far it's looking quite good at step 1600. It's starting to resemble the person I'm training.

mhirki:
First training run is complete. Final result looks pretty good. It looks a bit undertrained but that's kinda what I was expecting with Prodigy optimizer.

I uploaded the LoRA to Hugging Face in case someone wants to experiment with it:
https://huggingface.co/mikaelh/flux-sanna-marin-lora-v0.1

Step 100 validation image:

step_100_validation_(1024, 1024)

Step 2000 validation image:

step_2000_validation_(1024, 1024)

bghira:
interestingly there is some heavy patch artifacting going on there. let me demonstrate.

original model has it too. but it isn't as noticeable. this is something inherent to the DiT models that use PatchEmbeds.

image

using laplace edge detection we can really isolate the unnatural lines that are present throughout the original Flux outputs. for what it's worth, their Pro model does this too:

image
image
`a cloud of dust settling on a scene around a golden retriever dog`

the same kind of 16x16 grid artifacts. it manifests as colour and contrast aberrations. and it seems to be exacerbated issue when you don't use attention masking for the prompt embeds. Flux doesn't use attention masking.

I am curious if the issue disappears when you use 512 token limit for training Dev and 256 for Schall. this can be edited manually in the `text_embeds.py` file. you will have to re-cache them.

here is what your lora image looks like for comparison:
image

it's a lot more noisy but it is not all bad.

can you share more about your settings for this run? is it using DoRA?

bghira:
> A minor comment: usually int4 kernels tend to only work with bf16. Just a note to keep in mind.

which interaction are you referring to here? what's expected to be in bf16? fwiw Mac MPS has great bf16 support but kinda-crappy fp16 support, so really we're good here if that's what your concern was

mhirki:
> the same kind of 16x16 grid artifacts. it manifests as colour and contrast aberrations. and it seems to be exacerbated issue when you don't use attention masking for the prompt embeds. Flux doesn't use attention masking.

Yeah, the grid pattern also shows up very clearly in ComfyUI previews during inference. It doesn't bother me that much but I guess getting rid of it would be a nice bonus.

> I am curious if the issue disappears when you use 512 token limit for training Dev and 256 for Schall. this can be edited manually in the `text_embeds.py` file. you will have to re-cache them.

You mean increasing `max_sequence_length` from 256 to 512? I guess I could try that at some point.

> can you share more about your settings for this run? is it using DoRA?

It's a normal LoRA, not DoRA.

Here are all my config.env changes:
```
export FLUX=true
export LEARNING_RATE=1.0 #@param {type:"number"}
export MODEL_NAME="black-forest-labs/FLUX.1-dev"
export DEBUG_EXTRA_ARGS="--report_to=tensorboard"
export MAX_NUM_STEPS=2000
export BASE_DIR="/nvme/home/mikaelh/Stable_Diffusion/bghira/output"
export DATALOADER_CONFIG="${BASE_DIR}/multidatabackend_sanna_marin_flux.json"
export OUTPUT_DIR="${BASE_DIR}/models"
export VALIDATION_PROMPT="closeup of sanna marin"
export VALIDATION_NUM_INFERENCE_STEPS=20
export VALIDATION_RESOLUTION="1024x1024"
export TRAIN_BATCH_SIZE=1
export GRADIENT_ACCUMULATION_STEPS=1
export LR_WARMUP_STEPS=$((MAX_NUM_STEPS / 10))
export USE_XFORMERS=false
export OPTIMIZER="prodigy"
export TRAINER_EXTRA_ARGS="--base_model_precision=fp8-quanto --i_know_what_i_am_doing"
```

mhirki:
I updated to latest code from main branch and applied changes from this pull request on top of it. I'm now running an FP8 test run. It seems to be running a bit slower with speed currently hovering around 7.0s/it.

I also edited helpers/caching/text_embeds.py so that `max_sequence_length` is now 512. I verified that the cached text embeds are now 4 MB instead of 2 MB.

bghira:
use `--compress_disk_cache` if you want to reduce that 🌚

the sequence length fix should also be in this branch now.

bghira:
also, training with batch size more than 1 works.

a-l-e-x-d-s-9:
There is a way to load it with ComfyUI now?

mhirki:
> There is a way to load it with ComfyUI now?

Not yet. There's an issue for it: https://github.com/comfyanonymous/ComfyUI/issues/4202

sayakpaul:
> https://github.com/bghira/SimpleTuner/pull/622#issuecomment-2267377666

I saw int2 precision being used. But since you mentioned bf16 isn't a problem, that should be covered. What I tried to say is that precisions like that only tend to work in quanto with bf16 kernels. We discuss that here: https://huggingface.co/blog/quanto-diffusers#how-about-int4.

mhirki:
FP8 training is complete and uploaded to the Hub:

https://huggingface.co/mikaelh/flux-sanna-marin-lora-v0.2-fp8

At step 2800 everything was still fine:

step_2800_validation_(1024, 1024)

At step 2900 the validation image blows up:

step_2900_validation_(1024, 1024)

At step 3600 a miracle recovery happened:

step_3600_validation_(1024, 1024)

Finally at step 4000 the composition changed:

step_4000_validation_(1024, 1024)

Annoyingly, the 16x16 grid pattern was clearly there at step 3600. The change of composition appears to have mostly hidden it in the final image.

FP8 version works better with ComfyUI because ComfyUI doesn't offer the option for int8 quantization.

Maybe I'll try training a LoRA on the Schnell model next.

sayakpaul:
Thank you for your updates! The results are very nice!

Some questions for you so that we can better document things in `diffusers`. 

* How should users load https://huggingface.co/mikaelh/flux-sanna-marin-lora-v0.2-fp8 in `diffusers`. 
* "Maybe I'll try training a LoRA on the Schnell model next." -- a little note that schnell is a timstep-distilled model so the training dynamics might differ a bit from the regular models. 
* "FP8 version works better with ComfyUI because ComfyUI doesn't offer the option for int8 quantization." -- is it a ComfyUI specific thing or did you mean to compare any implementation differences between Comfy and Diffusers?

mhirki:
> * How should users load https://huggingface.co/mikaelh/flux-sanna-marin-lora-v0.2-fp8 in `diffusers`.

I wrote a simple inference script. I just pushed it to the Hub as well.

```
import torch
from diffusers import FluxPipeline

# Loading pipeline directly as FP8 doesn't seem to work
#pipe = FluxPipeline.from_pretrained("black-forest-labs/FLUX.1-dev", torch_dtype=torch.float8_e4m3fn)

pipe = FluxPipeline.from_pretrained("black-forest-labs/FLUX.1-dev", torch_dtype=torch.bfloat16)

pipe.load_lora_weights("mikaelh/flux-sanna-marin-lora-v0.2-fp8", weight_name="pytorch_lora_weights.safetensors")

# Quantization is slow but necessary if VRAM is limited to 24 GB
if 1:
    from optimum.quanto import freeze, qfloat8, qint8, quantize

    weight_quant = qfloat8

    # Quantize transformer and text encoder similar to SimpleTuner
    quantize(pipe.transformer, weights=weight_quant)
    freeze(pipe.transformer)

    quantize(pipe.text_encoder, weights=weight_quant)
    freeze(pipe.text_encoder)

    quantize(pipe.text_encoder_2, weights=weight_quant)
    freeze(pipe.text_encoder_2)

pipe.enable_model_cpu_offload()

prompt = "closeup of sanna marin"
out = pipe(
    prompt=prompt,
    guidance_scale=3.5,
    height=1024,
    width=1024,
    num_inference_steps=20,
).images[0]
out.save("image.png")
```

Highest VRAM usage I saw was around 15.4 GB with this script. It's also quite easy to make int8 variation out of it.

> * "Maybe I'll try training a LoRA on the Schnell model next." -- a little note that schnell is a timstep-distilled model so the training dynamics might differ a bit from the regular models.

Yes, I'm expecting problems but we'll see what happens.

> * "FP8 version works better with ComfyUI because ComfyUI doesn't offer the option for int8 quantization." -- is it a ComfyUI specific thing or did you mean to compare any implementation differences between Comfy and Diffusers?

This comment was strictly about the current state of ComfyUI. I tried both int8 and fp8 versions in ComfyUI. FP8 was working and int8 wasn't really working when combined with a model loaded with float8_e4m3fn precision.

sayakpaul:
Thank you so much for providing the additional details.

bghira:
i think comfy needs a quanto node

a-l-e-x-d-s-9:
@mhirki 
I tested [flux-sanna-marin-lora-v0.2-fp8 3900 step](https://huggingface.co/mikaelh/flux-sanna-marin-lora-v0.2-fp8/blob/main/checkpoint-3900/pytorch_lora_weights.safetensors) with ComfyUI.
With Dev version, by rows:
1. Without Lora - no trigger word prompt: woman face centered, white shirt, simple background
2. Without Lora - with trigger word prompt: sanna marin face centered, white shirt, simple background
3. With Lora - no trigger word prompt: woman face centered, white shirt, simple background
4. With Lora - with trigger word prompt: sanna marin face centered, white shirt, simple background

sanna_marin_lora_on_dev_st3900_1-without-lora-no-trigger_2-without-lora-with-trigger_3-with-lora-no-trigger_4-with-lora-with-trigger

With Schnell version, by rows:
1. Without Lora - no trigger word prompt: woman face centered, white shirt, simple background
2. Without Lora - with trigger word prompt: sanna marin face centered, white shirt, simple background
3. With Lora - no trigger word prompt: woman face centered, white shirt, simple background
4. With Lora - with trigger word prompt: sanna marin face centered, white shirt, simple background

sanna_marin_lora_on_schnell_st3900_1-without-lora-no-trigger_2-without-lora-with-trigger_3-with-lora-no-trigger_4-with-lora-with-trigger

sayakpaul:
Could be nice to explore too: https://huggingface.co/sayakpaul/FLUX.1-merged

image

mhirki:
@a-l-e-x-d-s-9
Yeah, I didn't have any regularisation so the LoRA is gonna produce that face always regardless of what the prompt says.

I'm surprised that my LoRA is working with the Schnell model. My current attempt at training a LoRA directly on the Schnell model isn't looking good.

a-l-e-x-d-s-9:
@mhirki 
I don't think you need regularization images, when you use the Lora you want to have the full effect without prompting for it too much.

bghira:
speed of loading on m3 max:

```
2024-08-05 07:21:09,991 [INFO] (helpers.training.quantisation) Loading Quanto for LoRA training. This may take a few minutes.
2024-08-05 07:21:09,991 [INFO] (helpers.training.quantisation) Quantising FluxTransformer2DModel. Using int8-quanto.
2024-08-05 07:21:38,767 [INFO] (helpers.training.quantisation) Freezing model.
2024-08-05 07:21:49,001 [INFO] (helpers.training.quantisation) Quantising CLIPTextModel. Using int8-quanto.
2024-08-05 07:21:49,269 [INFO] (helpers.training.quantisation) Freezing model.
2024-08-05 07:21:49,413 [INFO] (helpers.training.quantisation) Quantising T5EncoderModel. Using int8-quanto.
2024-08-05 07:21:50,358 [INFO] (helpers.training.quantisation) Freezing model.
2024-08-05 07:21:50,810 [INFO] (__main__) Using LoRA training mode (rank=1)
```

mhirki:
> @mhirki I don't think you need regularization images, when you use the Lora you want to have the full effect without prompting for it too much.

I guess that depends on what exactly you're prompting for. If you want to put two people into the prompt, then a LoRA effecting every face is a problem. So it depends on the use case.

The LoRA is trained on 128 images so it should have plenty of flexibility in single subject compositions.

bghira:
for PixArt and SD3, we're using `--max_grad_norm=0.01` hard limited because those two DiT prefer to have smaller gradient scales.

in empirical research we found training SD3 from scratch requires a higher max_grad_norm value.

Flux is currently not limited - the `max_grad_norm` is likely running at the default limit of `2`.

if it's not too much trouble, you could try unleashing this by setting it to 0, and you can limit it to 0.01 if you are interested in having a more gradual, stable training run.

mhirki:
> speed of loading on m3 max:

Damn, Apple silicon is pretty damn fast. My Intel CPU @ 5 GHz spends about 68 seconds doing quantisation + freezing for the transformer.

```
2024-08-05 14:22:26,424 [INFO] (helpers.training.quantisation) Loading Quanto for LoRA training. This may take a few minutes.
2024-08-05 14:22:26,424 [INFO] (helpers.training.quantisation) Quantising FluxTransformer2DModel. Using fp8-quanto.
2024-08-05 14:22:26,424 [WARNING] (helpers.training.quantisation) An earlier experimental build of this code erroneously used int8 instead of fp8. If you are resuming training and see errors, please use int8 instead of fp8.
2024-08-05 14:23:16,184 [INFO] (helpers.training.quantisation) Freezing model.
2024-08-05 14:23:34,638 [INFO] (helpers.training.quantisation) Quantising CLIPTextModel. Using fp8-quanto.
2024-08-05 14:23:34,638 [WARNING] (helpers.training.quantisation) An earlier experimental build of this code erroneously used int8 instead of fp8. If you are resuming training and see errors, please use int8 instead of fp8.
2024-08-05 14:23:34,697 [INFO] (helpers.training.quantisation) Freezing model.
2024-08-05 14:23:34,735 [INFO] (helpers.training.quantisation) Quantising T5EncoderModel. Using fp8-quanto.
2024-08-05 14:23:34,735 [WARNING] (helpers.training.quantisation) An earlier experimental build of this code erroneously used int8 instead of fp8. If you are resuming training and see errors, please use int8 instead of fp8.
2024-08-05 14:23:34,789 [INFO] (helpers.training.quantisation) Freezing model.
```

Yeah, I do remember `--max_grad_norm=0.01` having a positive effect in the early days of SD3. Maybe it can rescue my Schnell LoRA training.

sayakpaul:
A nice exploration could be try to tune the merged model I shared above as it helps reduce the number of steps. 

@bghira the speed is alluring. But inference breaks no because of rope?

sayakpaul:
So the speed benefits seem to be prevalent for inference for now?

bghira:
yes, as long as someone is not training on Mac, they can still inference on pytorch 2.3.1 but i think they need to use an older version of Quanto for that, one which does not rely on py2.4

bghira:
in fact using the older quanto package on 2.3.1 at a rofl-worthy int2 level of precision i get 2 seconds per step for the dev model... amazing.

mhirki:
I'm making good progress on training the Schnell model. After some initial failures, I started reading the SD3-Turbo paper. Section 5 is particularly interesting:

> We train across four discrete timesteps t ∈ [1, 0.75, 0.5, 0.25].

In addition, they used a custom noise schedule:

> After 500 iterations, the focus shifts towards full noise (p = [0.7, 0.1, 0.1, 0.1]) to refine single-shot performance.

So I wrote my own hackjob implementation of this, diff below. It's a bit messy and I'm not sure if it's 100% correct but it's getting me some results.

```diff
diff --git a/train.py b/train.py
index c877fb7..6af072a 100644
--- a/train.py
+++ b/train.py
@@ -480,6 +480,13 @@ def main():
             args.pretrained_model_name_or_path, subfolder="scheduler"
         )
         noise_scheduler_copy = copy.deepcopy(noise_scheduler)
+        if "schnell" in args.pretrained_model_name_or_path.lower():
+            # 4-step noise schedule [0.7, 0.1, 0.1, 0.1] from SD3-Turbo paper
+            for i in range(0, 250):
+                noise_scheduler_copy.sigmas[i] = 0.7
+            for i in range(250, 1000):
+                noise_scheduler_copy.sigmas[i] = 0.1
+            print(f"sigmas: {noise_scheduler_copy.sigmas}")
 
     else:
         if args.legacy:
@@ -1823,9 +1830,15 @@ def main():
                     else:
                         u = torch.rand(size=(bsz,), device="cpu")
 
-                    indices = (
-                        u * noise_scheduler_copy.config.num_train_timesteps
-                    ).long()
+                    if "schnell" in args.pretrained_model_name_or_path.lower():
+                        # We need to train only timesteps [1, 0.75, 0.5, 0.25] based on SD3-Turbo paper
+                        quarter_step = int(noise_scheduler_copy.config.num_train_timesteps / 4)
+                        indices = ((u * 4).long() + 1) * quarter_step - 1
+                        #indices = (u * 4).long() * quarter_step - 1
+                    else:
+                        indices = (
+                            u * noise_scheduler_copy.config.num_train_timesteps
+                        ).long()
                     timesteps = noise_scheduler_copy.timesteps[indices].to(
                         device=latents.device
                     )
```
Validation image after 100 steps:

step_100_validation_(1024, 1024)

Another validation image after 600 steps:

step_600_validation_(1024, 1024)

The image is noisy as hell but it's clearly taking shape. I guess the obvious solution to the noise problem would be to increase batch size since I'm currently training at batch size of 1. But I'll just leave it running overnight for now.

bghira:
can you check the sigmas that you hit at inference time and compare them?

mhirki:
Here's a zoomed in view of training loss from Tensorboard. The loss should correlate with sigmas. It's jumping up roughly 25% of the time as intended.

Kuvakaappaus_20240806_011028_schnell_optimization_loss

mhirki:
Well, the final result at 4000 steps ended up having even more noise:

step_4000_validation_(1024, 1024)

Interestingly, I can get rid of most the noise by simply dropping LoRA weight to 0.5:

fluxs_1722929741_0001

Now I'm just left with a film grain effect which isn't that bad.

Apparently a Schnell LoRA doesn't really do much when combined with the Dev model. This makes sense because the Schnell LoRA is trying to do most of the work in the first step.

I will try increasing batch size for the next run.

mhirki:
Actually, I decided to fix my sigmas since they really should be going from 1.0 to 0. So my sigmas for the 4 steps are now 1.0, 0.3, 0.2 and 0.1.

```diff
diff --git a/train.py b/train.py
index c877fb7..87ac346 100644
--- a/train.py
+++ b/train.py
@@ -480,6 +480,18 @@ def main():
             args.pretrained_model_name_or_path, subfolder="scheduler"
         )
         noise_scheduler_copy = copy.deepcopy(noise_scheduler)
+        if "schnell" in args.pretrained_model_name_or_path.lower():
+            # 4-step noise schedule [0.7, 0.1, 0.1, 0.1] from SD3-Turbo paper
+            print(f"sigmas before: {noise_scheduler_copy.sigmas}")
+            for i in range(0, 250):
+                noise_scheduler_copy.sigmas[i] = 1.0
+            for i in range(250, 500):
+                noise_scheduler_copy.sigmas[i] = 0.3
+            for i in range(500, 750):
+                noise_scheduler_copy.sigmas[i] = 0.2
+            for i in range(750, 1000):
+                noise_scheduler_copy.sigmas[i] = 0.1
+            print(f"sigmas after: {noise_scheduler_copy.sigmas}")
 
     else:
         if args.legacy:
@@ -1823,9 +1835,15 @@ def main():
                     else:
                         u = torch.rand(size=(bsz,), device="cpu")
 
-                    indices = (
-                        u * noise_scheduler_copy.config.num_train_timesteps
-                    ).long()
+                    if "schnell" in args.pretrained_model_name_or_path.lower():
+                        # We need to train only timesteps [1, 0.75, 0.5, 0.25] based on SD3-Turbo paper
+                        quarter_step = int(noise_scheduler_copy.config.num_train_timesteps / 4)
+                        indices = ((u * 4).long() + 1) * quarter_step - 1
+                        #indices = (u * 4).long() * quarter_step - 1
+                    else:
+                        indices = (
+                            u * noise_scheduler_copy.config.num_train_timesteps
+                        ).long()
                     timesteps = noise_scheduler_copy.timesteps[indices].to(
                         device=latents.device
                     )
```

At 600 steps it seems to be working:

step_600_validation_(1024, 1024)

No more crazy amounts of noise. Somehow this looks like it's learning more slowly but maybe that's how Prodigy optimizer reacts to training loss having a huge variance.

sayakpaul:
A higher learning rate could beneficial for Prodigy. Have you tried it already?

My experiments also suggest that a higher beta2 in Adam is more stable. So, perhaps worth trying.

sayakpaul:
> i need to check the model configs again to see if the proper values are in there

`max_sequence_length` isn't a model config var. The source of truth for this determination is the original inference code from BFL.

sayakpaul:
Yeah not a bad idea.

bghira:
can also update DF-IF's to 77 and PixArt's to 300? wdyt @sayakpaul

bghira:
@sayakpaul do you have any ideas for a schnell sampler that might do this more cleanly and be reusable by other projects than simpletuner? eg. in diffusers directly

sayakpaul:
Sampler? The schnell sampler is already different in the sense that it doesn't use dynamic shifting of scales whereas the dev sampler use that. This is evident from the `config.json`s present in the respective `scheduler` directories.

bghira:
well, scheduler, since they're combined in Diffusers. there's a code change above that essentially splits the 1000 timestep schedule into 4 pieces and forces the sigma at each chunk to the same value. so you end up with 4 different sigmas that may be obtained from the 1000 steps.

sayakpaul:
I got confused because the question was asked just after the sequence length thing. 

Actually it’s better done in the training code and sigmas are sampled in the training code when required. I remember Daniel did something similar for ADD. 

In the case of inference, the computation of sigmas happens within the scheduler which seems to be different from what we are doing here. So, not sure about the best way to accommodate the changes in the scheduler. But having said that, I would not mind adding a utility to our training utils. Would that work?

bghira:
hmm, to me, it's of utmost importance to align training with inference. if we are training schnell on just 4 sigmas but inference is capable of handling others, are we really aligned?

sayakpaul:
The original inference code doesn’t really do anything special about that. That is our source of truth in diffusers. So, I am not sure. That is why I suggested the inclusion of a util in training.

bghira:
alright. that is built in now. you can use `--flux_fast_schedule` option to enable. thank you!

Kaznakoff (approved):

PR #629, opened 2024-08-04
Flux: Fix random validation errors due to some tensors being on the cpu

mhirki:
Flux had random validation errors similar to SD3 in issue #546. This pull request makes sure all tensors are moved to the GPU before calling the transformer. Code is mostly copied from helpers/sd3/pipeline.py. In the case of Flux, the transformer also takes in `txt_ids` and `img_ids` parameters which also needed to be moved to GPU.

I did some testing with extra debug prints and the validation image was successfully generated even when tensors were originally on the CPU.


bghira:
thanks!

PR #645, opened 2024-08-05
csv backend updates

williamzhuk:
- hashing instead of shortening
- csv.py renamed to csv_.py to avoid conflict with pandas internal csv.py


helpers/data_backend/csv_.py
```diff
@@ -99,8 +94,13 @@ def read(self, location, as_byteIO: bool = False):
...
         if not location.startswith("http"):
             # read from local file
-            with open(location, "rb") as file:
-                data = file.read()
+            hashed_location = path_to_hashed_path(location, hash_filenames=self.hash_filenames and not already_hashed)
+            try:
+                with open(hashed_location, "rb") as file:
+                    data = file.read()
+            except FileNotFoundError as e:
+                print(f'ask was for file {location} bound to {hashed_location}')
```
bghira: can you use tqdm.write or logger.error ?

helpers/data_backend/csv_.py
```diff
@@ -279,9 +279,9 @@ def torch_save(self, data, location: Union[str, Path, BytesIO]):
         """
         Save a torch tensor to a file.
         """
+
         if isinstance(location, str) or isinstance(location, Path):
-            if location not in self.df.index:
-                self.df.loc[location] = pd.Series()
+            location = path_to_hashed_path(location, self.hash_filenames)
```
bghira: can we update this method to use clone() before saving, like the local backend does?

helpers/data_backend/factory.py
```diff
@@ -1,6 +1,6 @@
 from helpers.data_backend.local import LocalDataBackend
 from helpers.data_backend.aws import S3DataBackend
-from helpers.data_backend.csv import CSVDataBackend
+from helpers.data_backend.csv_ import CSVDataBackend
```
bghira: i would prefer we call it csv_url_list if we're changing it

bghira (changes requested):
thanks for the updates, hashing things makes sense. but i'd like the comments resolved first 👍

helpers/data_backend/factory.py
bghira:
```suggestion
from helpers.data_backend.csv_url import CSVDataBackend
```

helpers/data_backend/csv_.py
bghira:
```suggestion
                tqdm.write(f'ask was for file {location} bound to {hashed_location}')
```

bghira:
i updated it to use file extensions instead of str_pattern to search. it still works. @williamzhuk can you follow-up with the other changes?

bghira (approved):

PR #647, opened 2024-08-05
quanto: allow further vram reduction with bf16 base weights, reorder model loading operations to evacuate text encoders before DiT loads

bghira:
this is kind of an experimental change.

i was investigating how to fix the LoRA loading issue when we try and resume a checkpoint, as that is currently broken upstream in Quanto.

however, i also discovered that if we send the model to its target dtype before quantising it, the weight is reduced and we can train in even less memory.

this requires the adam_bf16 optimiser be used, so the error messages are updated to reflect that.

you can still use adafactor and others, but it's noted as a bit heavier and requires fp32 base default dtype.

### DIFF
--- a/config/config.env.example
+++ b/config/config.env.example
@@ -5,6 +5,8 @@
 # full - requires lots of vram, trains very slowly, needs a lot of data and concepts.
 export MODEL_TYPE='lora'
 
+# SDXL is trained by default, but you will need to enable one of these options for anything else.
+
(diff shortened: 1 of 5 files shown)


bghira:
on my m3 max,

**quantised using int8**
- the previous code used ~32.5GB to train at rank-16, bsz=1
- this code uses ~24.5GB to train at rank-16, bsz=1

pure bf16 training is the same as before, ~32G

PR #661, opened 2024-08-07
refactor saving utitilites part ii

sayakpaul:


helpers/training/save_hooks.py
```diff
@@ -150,12 +169,6 @@ def _save_lora(self, models, weights, output_dir):
                 text_encoder_2_lora_layers_to_save = convert_state_dict_to_diffusers(
                     get_peft_model_state_dict(model)
                 )
-            # elif isinstance(
-            #     model, type(unwrap_model(self.accelerator, self.text_encoder_3))
-            # ):
-            #     text_encoder_3_lora_layers_to_save = convert_state_dict_to_diffusers(
-            #         get_peft_model_state_dict(model)
-            #     )
```
sayakpaul, line 158: text encoder 3 lora training is anyway not supported so, I have deleted `text_encoder_3` from the face of `save_hooks.py`. LMK if it should have been done differently.

helpers/training/save_hooks.py
```diff
@@ -173,29 +186,27 @@ def _save_lora(self, models, weights, output_dir):
                 weights.pop()
 
         if self.args.flux:
-            FluxPipeline.save_lora_weights(
+            self.pipeline_class.save_lora_weights(
```
sayakpaul, line 189: Determine the pipeline class only once and reuse.

helpers/training/save_hooks.py
```diff
@@ -123,6 +130,18 @@ def __init__(
...
 
+        self.pipeline_class = None
+        if self.args.sd3:
+            self.pipeline_class = PIPELINE_MAPPING["sd3"]
+        elif self.args.legacy:
+            self.pipeline_class = PIPELINE_MAPPING["legacy"]
+        elif self.flux:
+            self.pipeline_class = PIPELINE_MAPPING["flux"]
+        else:
+            self.pipeline_class = PIPELINE_MAPPING["sdxl"]
```
bghira, line 141: hmm i guess we're relying on the sdxl pipeline for kolors but what about pixart?

bghira: oh, right - pixart doesn't do loras

bghira (approved):

Issue #656, opened 2024-08-06
Flux Lora training crashed on the first save.


a-l-e-x-d-s-9:
Flux Lora training crashed on the first save.
Error log:
``` log
Epoch 1/22, Steps:   3%|█                                        | 40/1500 [30:37<18:40:40, 46.06s/it, lr=1, step_loss=0.191]

[2024-08-06 16:42:19,724] [INFO] [real_accelerator.py:158:get_accelerator] Setting ds_accelerator to cuda (auto detect)
cannot import name 'log' from 'torch.distributed.elastic.agent.server.api' (/workspace/SimpleTuner/.venv/lib/python3.10/site-packages/torch/distributed/elastic/agent/server/api.py)
Traceback (most recent call last):
  File "/workspace/SimpleTuner/train.py", line 2765, in <module>
    main()
  File "/workspace/SimpleTuner/train.py", line 2412, in main
    accelerator.save_state(save_path)
  File "/workspace/SimpleTuner/.venv/lib/python3.10/site-packages/accelerate/accelerator.py", line 2955, in save_state
    hook(self._models, weights, output_dir)
  File "/workspace/SimpleTuner/helpers/training/save_hooks.py", line 249, in save_model_hook
    self._save_lora(models=models, weights=weights, output_dir=output_dir)
  File "/workspace/SimpleTuner/helpers/training/save_hooks.py", line 137, in _save_lora
    if isinstance(model, type(unwrap_model(self.accelerator, self.unet))):
  File "/workspace/SimpleTuner/helpers/training/wrappers.py", line 5, in unwrap_model
    model = accelerator.unwrap_model(model)
  File "/workspace/SimpleTuner/.venv/lib/python3.10/site-packages/accelerate/accelerator.py", line 2515, in unwrap_model
    return extract_model_from_parallel(model, keep_fp32_wrapper)
  File "/workspace/SimpleTuner/.venv/lib/python3.10/site-packages/accelerate/utils/other.py", line 80, in extract_model_from_parallel
    from deepspeed import DeepSpeedEngine
  File "/workspace/SimpleTuner/.venv/lib/python3.10/site-packages/deepspeed/__init__.py", line 22, in <module>
    from . import module_inject
  File "/workspace/SimpleTuner/.venv/lib/python3.10/site-packages/deepspeed/module_inject/__init__.py", line 6, in <module>
[...]

PR #666, opened 2024-08-07
merge: reorder model casting, fix kolors on newer diffusers, update cuda, refactor save hooks

bghira:
this will wait a bit longer to shake issues out before merging


bghira:
Resolves #656

PR #694, opened 2024-08-09
Flux: Purge text encoders from system RAM.

mhirki:
Get rid of FluxPipeline inside TextEmbeddingCache that is keeping text encoders in system RAM. This frees up RAM for other things like quantization. Peak RAM usage during fp8 quantization is about 45.4 GB after this change.


bghira:
yeah, i guess it's really not needed after the preprocessing stage. thank you.

PR #701, opened 2024-08-10
Update Dockerfile

FKouhai:
removed uneeded tools for a container,arguably ssh isnt needed either since theres the possibility to docker exec into the container,vim isnt needed since you would have a volume for the container mapped to the host and you can edit files there

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,21 +16,12 @@ RUN apt-get install -y --no-install-recommends openssh-server \
                                                openssh-client \
                                                git \
                                                git-lfs \
-                                               wget \
                                                curl \
-                                               tmux \
-                                               tldr \
-                                               nvtop \
-                                               vim \
                                                rsync \
-                                               net-tools \
                                                less \
-                                               iputils-ping \
                                                7zip \
                                                zip \
-                                               unzip \
-                                               htop \
-                                               inotify-tools
+                                               unzip
 
 # Set up git to support LFS, and to store credentials; useful for Huggingface Hub
 RUN git config --global credential.helper store && \


bghira:
doesn't this stop using ssh on eg. Runpod? cc @komninoschatzipapas

FKouhai:
both ssh packages were kept,I made the suggestion that ssh in a container is not needed and its arguably a bad idea,for a myriad of security reasons

bghira:
👍 Komninos is maintainer of that, so I defer to them

komninoschatzipapas:
@FKouhai I think your thoughts are right for users that'd run the Dockerfile on their local machine. However, as bghira pointed out, Runpod, Vast, and other cloud services don't allow you to access the Docker host of the container.

If we were to remove SSH, there would be no way for users to access the container on these cloud services.

I'm happy to hear any security improvements we can make on that front. I'd generally caution against deploying this container without setting an SSH key (both Runpod & Vast do this automatically.)

I'm also aware of the fact that some of the packages I added are not necessarily needed to use SimpleTuner. However, I've found that more of than not, things in ML bug out. And having all these tools makes things more convenient when that happens.

FKouhai:
I think Id be better to have it being a web api or just not a container,having ssh and a lot of those tools kind of goes against the nature of what containers are since they are ephemeral it doesnt make too much sense to have those tools in place, if the end goal for this is being a web API I wouldnt mind helping out with that part,in general granting shell access to a container specially with docker where theres been a lot of CVE's that allowed an attacker to exit the container and leverage access to the host itself I dont think it would be wise

bghira:
can you show an example of how the docker container can be taken over through the availability of SSH? this sounds like a serious concern and i would like to resolve it, but i need to understand the vector first

bghira:
re: containers being cattle and not kittens, this is a somewhat antiquated idea from about 10 years ago where people thought microservices were the future.

but since then, RunPod, Vast, Lambda, Coreweave, TensorDock, all of these companies are selling access to cloud machines through the use of containers that run a full Linux environment as a lower-overhead replacement for virtualisation.

try logging into one and you'll see what I mean

FKouhai:
https://www.paloaltonetworks.com/blog/prisma-cloud/leaky-vessels-vulnerabilities-container-escape/

Im sorry but containers are ephemeral by nature they are not VMs its not an antiquated idea it is what they are, you can access containers via web API, in case you want to have a shell in your container (docker exec is a thing and so is kubectl exec),as I said its not only a bad idea its just not what containers are for,I could give you a myriad of companies that have microservices and their pods/containers are mainly for API(s) or web access

bghira:
this is a philosophical concept, not any strict technical limitation on containers.

if you really want to get into "what they are", it's just a fancy chroot.

PR #707, opened 2024-08-10
Add the ability to train all nn.Linear for flux

AmericanPresidentJimmyCarter:


train.py
```diff
@@ -969,6 +969,19 @@ def main():
                         "add_v_proj",
                         "to_add_out.0",
                     ]
+                elif args.flux_lora_target == "full":
```
bghira:
```suggestion
                elif args.flux_lora_target == "all+norms":
```

AmericanPresidentJimmyCarter: It's not norms, it's ff. I'll change it to "all+ffs"

bghira (approved):

PR #711, opened 2024-08-10
flux: import some of kohya suggestions (wip)

bghira:


train.py
```diff
@@ -500,7 +500,9 @@ def main():
         from diffusers import FlowMatchEulerDiscreteScheduler
 
         noise_scheduler = FlowMatchEulerDiscreteScheduler.from_pretrained(
-            args.pretrained_model_name_or_path, subfolder="scheduler"
+            args.pretrained_model_name_or_path,
+            subfolder="scheduler",
+            shift=1 if args.flux else 3,
```
bghira, line 505: well this would be if we use timestep_scheme=sd3 for flux. but we don't. so this does nothing

mhirki:
The new --timestep_scheme=flux seems to be equivalent to --weighting_scheme=logit_normal which was also used by SD3 so the naming of the options is kinda confusing. It is a good idea to have a separate --timestep_scheme option because --weighting_scheme was also confusing because it could either change timestep sampling or loss conditioning.

We should perhaps have --timestep_scheme option with possible values logit_normal, mode and uniform. Then the terminology would be in line with the SD3 paper.

bghira:
it isnt though, the timestep selection is based on sigmoid and the prev example derived the sigmoids from `t` instead of the other way around. i agree the naming and options need a final pass cleanup which i will do this evening after some much needed food

mhirki:
Yes, the previous implementation used 1000 timestep buckets and the new implementation is fully continuous which is better obviously. Logit-normal distribution vanishes at sigma=0 and sigma=1 so I think the new implementation is better near those locations.

bghira:
i am probably going to get rid of most of the SD3-specific stuff, because it was all experimental even in the paper, and we now have better results.

PR #723, opened 2024-08-11
Add input perturbation

mhirki:
Input perturbation is a simple trick to speed up training convergence. We just add a little bit of extra noise to the inputs fed to the model.

Input perturbation has already been added to diffusers before:
https://github.com/huggingface/diffusers/issues/3293

Kohya calls this --ip_noise_gamma:
https://github.com/kohya-ss/sd-scripts/pull/798

The original research paper is here:
https://arxiv.org/pdf/2301.11706

I tested input perturbation with Flux LoRA training and it seems to work quite well. I did a little test run at 512px resolution and adamw_bf16 optimizer with 4e-4 learning rate.

500 steps without input perturbation:
step_500_validation_(512, 512)

500 steps with input perturbation set to 0.1:
step_500_validation_(512, 512)

I'd say the latter image is already starting look almost fully converged at just 500 steps.

Also, there seems to be less bleeding into the unconditional validation images.

### DIFF
--- a/OPTIONS.md
+++ b/OPTIONS.md
@@ -1312,7 +1312,8 @@ options:
                         work with flow-matching (e.g. SD3 and Flux).
   --input_perturbation_steps INPUT_PERTURBATION_STEPS
                         Only apply input perturbation over the first N steps
-                        with linear decay.
+                        with linear decay. This should prevent artifacts from
+                        showing up in longer training runs.
   --lr_end LR_END       A polynomial learning rate will end up at this value
                         after the specified number of warmup steps. A sine or
                         cosine wave will use this value as its lower bound for
--- a/OPTIONS.md
+++ b/OPTIONS.md
@@ -393,7 +393,9 @@ usage: train.py [-h] [--snr_gamma SNR_GAMMA] [--use_soft_min_snr]
                 [--sdxl_refiner_uses_full_range]
                 [--caption_dropout_probability CAPTION_DROPOUT_PROBABILITY]
                 [--delete_unwanted_images] [--delete_problematic_images]
-                [--offset_noise] [--lr_end LR_END] [--i_know_what_i_am_doing]
+                [--offset_noise] [--input_perturbation INPUT_PERTURBATION]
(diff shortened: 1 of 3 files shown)


mhirki:
I got almost the same result with int8 quantization and 0.1 for input perturbation:
step_500_validation_(512, 512)

bghira:
already had it once upon a time, but it's also really terribly performing on longer runs. feel free to experiment some more, but i don't want to really add it back in just yet.. or at all, really. i dislike the artifacting that it results in.

mhirki:
One idea might be to keep input perturbation enabled only for the first N steps. Then it would only be used to speed up early convergence. Of course, such a setting can get tricky with warmup steps.

bghira:
could you propose a torch scheduler for input perturbations?

mhirki:
I think we could have linear decay for input perturbation. Torch has LinearLR but that hooks into an optimizer so I guess it's easier to roll my own implementation based on `global_step` at least for testing.

I did leave the int8 training running for 6000 steps to see if the LoRA would fry itself. So far it doesn't look that bad. There's maybe a hint of overfitting but this thing seems surprisingly resilient. Next run will be 6000 steps with input perturbation decaying over first 2000 steps.
step_6000_validation_(512, 512)

mhirki:
Second int8 training run with 0.1 input perturbation combined with linear decay over 2000 steps has finished.

Validation image at 500 steps still shows fast convergence:
step_500_validation_(512, 512)

And the final validation image at 6000 steps looks pretty much the same:
step_6000_validation_(512, 512)

Unconditional validation images have more variety in them after the input perturbation has decayed off. So I guess this way it ends up slightly less burned in.

bghira:
would you mind updating OPTIONS.md to explain this one?

mhirki:
Alright, I updated output of `python train.py -h` in OPTIONS.md to include input perturbation. I also rebased my commits just to make sure they will still merge with the main branch.

bghira:
not avoiding/ignoring this but been working on quanto resumption issues and other features needed urgently like pixel_area mode 👍 if i don't get to it this evening, i will tomorrow

PR #739, opened 2024-08-12
refactor train.py to reduce LoCs.

sayakpaul:
This is the first part and aims to improve the readability of `train.py` by:

* Separating chore LoCs into respective modules so that users can focus on more core things. 
* Since the chore LoCs now have their own modules, they should now be fairly extensible i.e., a new error check.  

With this approach, we are able to reduce about 421 LoC to the main entrypoint of SimpleTuner.


bghira:
tested. looks good, ty

sayakpaul:
Will drop a part II soon. Thanks for testing.

PR #752, opened 2024-08-13
Fixes #745 - do not throw last partial batch

### LINKED ISSUE #745: Do not throw away images if your dataset is not divisible by batch size
There has to be some reasonable way to do this.

burgalon:


bghira:
but did you test it with multiple gpus?

burgalon:
Yap

bghira:
can you share the details

bghira:
for example if we have a batch size of 4, and 1 image in the set, what happens?

burgalon:
in the case of the dataset of 19 images, it seemed to use all the 19 images, and i'm guessing it filled the batches using torch.choice properly? Is there anything specific you'd like me to check?

bghira:
the problem will be that the batch size must be met by the aspect sampler. or it will never yield anything

burgalon:
in my case i was using with:
```
    "crop": true,
    "crop_aspect": "square",
    "crop_style": "center",
```
but I'm not sure it matters for this case. The training seemed to go well and it seemed like all images were used as much as I can tell....

PR #756, opened 2024-08-14
add pixel_area resolution type

bghira:


helpers/image_manipulation/training_sample.py
```diff
@@ -157,10 +158,27 @@ def _set_resolution(self):
...
+            elif self.resolution == 1.0:
+                # update for 1024x1024
+                self.resolution = 1.05
+            elif self.resolution == 2.0:
+                # update for 1536x1536
+                self.resolution = 2.36
+            elif self.resolution == 4.0:
+                # update for 2048x2048
+                self.resolution = 4.19
             self.target_area = self.resolution * 1e6  # Convert megapixels to pixels
```
Beinsezii: I feel a simple mapping here would be better to provide more accurate values and self-document
```suggestion
        elif self.resolution_type == "area":
            # Convert pixel area to megapixels, remapping commonly used round values
            # to their pixel_area equivalents for compatibility purposes.
            self.target_area = {
                0.25: 512**2,
                0.5: 768**2,
                1.0: 1024**2,
                2.0: 1536**2,
                4.0: 2048**2,
            }.get(self.resolution, self.resolution * 1e6)
```

helpers/arguments.py
```diff
@@ -774,12 +774,13 @@ def parse_args(input_args=None):
...
         type=str,
-        default="area",
-        choices=["pixel", "area"],
+        default="pixel_area",
+        choices=["pixel", "area", "pixel_area"],
         help=(
             "Resizing images maintains aspect ratio. This defines the resizing strategy."
-            " If 'pixel', the images will be resized to the resolution by pixel edge."
+            " If 'pixel', the images will be resized to the resolution by the shortest pixel edge, if the target size does not match the current size."
             " If 'area', the images will be resized so the pixel area is this many megapixels."
```
Beinsezii: I believe this should be transparent about the 'silent' overrides. Something like?
```suggestion
            " If 'area', the images will be resized so the pixel area is this many megapixels. Common rounded values such as `0.5` and `1.0` will be implicitly adjusted to their squared size equivalents."
```

PR #773, opened 2024-08-15
add `ai-toolkit` option in --flux_lora_target choices

benihime91:
"ai-toolkit" choice was added to adapters for flux but not exposed in `--flux_lora_target` arguments. This PR fixes that


bghira:
my gosh, what an oversight. thank you.

PR #785, opened 2024-08-16
wip: optimi

bghira:
it doesn't work yet. this is for other developers to review and test.

- [x] import optimi configs as defaults
- [x] map old --parameters to their new args
- [x] removed adafactor
- [x] removed dadaptation
- [x] removed prodigy
- [x] recommend adafactor -> stableadamw
- [x] enable gradient release by default on all optims
- [x] handle training loop changes eg. remove adafactor-specific avoidance of gradient clipping in favour of optimi-aware decision
- [x] update model card logic to reference correct optimiser and config
- [x] update documentation to indicate changes to optimiser configurations
- [x] run memory benchmarks to measure savings


bghira:
cc @sayakpaul

bghira:
nope, it's really not going to work. quanto just doesn't work with Optimi

bghira:
not only do they not work with quanto, they also seem to require autocast despite claiming not to

bghira:
i figured out how to make it work with quanto and now rank 128 training consumes same memory as rank 4 training, so that's cool

Issue #788, opened 2024-08-16
Fail when the user uses the same image directory and the cache_dir is the same in dataset backends


AmericanPresidentJimmyCarter:
Prevents crash where the trainer attempts to mix shapes in a batch.

PR #794, opened 2024-08-17
next release

bghira:


bghira:
i kinda want #788 but i don't know whether there's any major issues people are impacted by first.

Issue #780, opened 2024-08-16
Error on FLUX.1 training


shencuifeng:
In the FLUX.1 training, when I set batch size greater than 1, and set **args.flux_fast_schedule** to true, it has this error on line 1479:
`RuntimeError: The size of tensor a (4) must match the size of tensor b (2) at non-singleton dimension 3`
Is this because missing `sigmas = sigmas.view(-1, 1, 1, 1)` on line 1473?

bghira:
are you training Schnell? this feature hasn't really been updated since the Flux changes started rolling in.

PR #803, opened 2024-08-18
Fixes for --flux_fast_schedule

mhirki:
This fixes issue #780. I tested it with a batch size of 2. I'm also fixing the schedule to what I believe is correct based on SD3-Turbo paper. Schnell training is still broken based on a quick 500 step test run. So the advice is still to stay away from Schnell.

### DIFF
--- a/train.py
+++ b/train.py
@@ -1551,16 +1551,15 @@ def main():
                             args.flow_matching_sigmoid_scale
                             * torch.randn((bsz,), device=accelerator.device)
                         )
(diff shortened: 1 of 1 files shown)


bghira:
would you be interested in attempting to port the assistant lora code from ostris' ai-toolkit?

mhirki:
Not really interested and I don't have enough spare time either.

bghira:
same :D

PR #806, opened 2024-08-18
Add attention masking to the custom helpers/flux/transformer.py

AmericanPresidentJimmyCarter:
- Fixes bugs in the Transformer2D class for flux
- Adds attention masking and the new FluxTransformer2DModelWithMasking model


bghira:
**note**: To others reviewing this code for adaptation or inclusion in their project, the license of this is **AGPLv3** and requires clean-room reimplementation to be useful via an incompatible license such as Apache2 or MIT.

PR #896, opened 2024-08-27
fix: make sure wandb_logs is always defined

ErwannMillon:
wandb_logs is undefined if `accelerator.sync_gradients` is False


bghira:
this one is incorrect. you must instantiate it above with the rest of `= []` options where the epoch begins.

PR #903, opened 2024-08-30
Update TRAINER_EXTRA_ARGS for model_family

barakyo:
## Background

In https://github.com/bghira/SimpleTuner/commit/79ded80af4803a88b71b484b3cefe8f9976b6d0d the `--flux`, `--sd3`, etc arguments were remove and instead need to be passed as `--model_family=<model>`. The `train.sh` script wasn't updated so if you're freshly cloning the repo and running the script it will fail to run.


bghira:
the intention is that people switch to `MODEL_FAMILY=sd3` etc, which is outlined in the config.env.example

PR #905, opened 2024-08-31
Fix Flux schedule shift and add resolution-dependent schedule shift

mhirki:
This is based on discussion after PR #892. This fixes --flux_schedule_shift by applying the shift to the sigmas. This makes it equivalent with Kohya's code.

This pull request also adds new option --flux_schedule_auto_shift for resolution-dependent shifting based on official Flux inference code:
https://github.com/black-forest-labs/flux/blob/b4f689aaccd40de93429865793e84a734f4a6254/src/flux/sampling.py#L66

People on Discord have been trying this yesterday already and results have been positive. Previous version of the code is here (before rebasing it due refactored train.py):
https://github.com/mhirki/SimpleTuner/commit/7cafb8677543923d11dd00197f0f122d69920cf4

I tested the code using my previous configuration from this run:
https://huggingface.co/mikaelh/flux-sanna-marin-v0.4-fp8-adan2

One of the observations from my own testing was that learning rate needed to be reduced. Initially I tried running with --flux_schedule_shift=3.1582 and lowering learning rate by 2x but that ended up overfitting so I ended up using a checkpoint from 2000 steps. Then I tried --flux_schedule_auto_shift with 5x lower learning rate and it turned out well.

| --flux_schedule_shift=3.1582 with 2x lower LR @ 2000 steps | --flux_schedule_auto_shift with 5x lower LR @ 3000 steps |
| --- | --- |
| flux_1725001441_0001 | flux_1725092497_0001 |
| flux_1725001553_0001 | flux_1725092526_0001 |
| flux_1725001583_0001 | flux_1725092556_0001 |
| flux_1725001613_0001 | flux_1725092586_0001 |
| flux_1725001643_0001 | flux_1725092624_0001 |
| flux_1725001674_0001 | flux_1725092663_0001 |
| flux_1725001704_0001 | flux_1725092693_0001 |
| flux_1725001735_0001 | flux_1725092723_0001 |

### DIFF
--- a/helpers/training/trainer.py
+++ b/helpers/training/trainer.py
@@ -117,9 +117,7 @@ from helpers.models.flux import (
     pack_latents,
     unpack_latents,
     get_mobius_guidance,
-)
-from diffusers.pipelines.flux.pipeline_flux import (
-    calculate_shift as calculate_shift_flux,
+    apply_flux_schedule_shift,
 )
 
 is_optimi_available = False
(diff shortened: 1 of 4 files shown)


helpers/training/trainer.py
```diff
@@ -1625,6 +1628,32 @@ def train(self):
...
+                                    self.noise_scheduler.config.max_shift,
+                                )
+                                training_logger.debug(
+                                    f"image_seq_len = {image_seq_len}, mu = {mu}, noise.shape = {noise.shape}"
+                                )
+                                flux_shift_value = math.exp(mu)
+                            if flux_shift_value is not None:
+                                sigmas = (sigmas * flux_shift_value) / (
+                                    1 + (flux_shift_value - 1) * sigmas
+                                )
```
bghira: let's go ahead and add this to a specific helper inside `helpers/models/flux/__init__.py` as i'm going to be removing a lot of the branching logic from trainer.py eventually anyway

bghira (changes requested):

helpers/training/trainer.py
mhirki: That's a good idea. I just implemented it.

PR #906, opened 2024-08-31
Fix text encoder nuking regression

mhirki:
Fix text encoders staying in system RAM after train.py was refactored in #899. This reduces system RAM usage by 9.1 GB for me while training Flux.


bghira:
god i should start paying you for finding these... thank you

PR #936, opened 2024-09-03
Fix caption_with_cogvlm.py  for cogvlm2 + textfile strategy

burgalon:


bghira:
can you use the following `.git/hooks/pre-commit` and `chmod +x` it:

```bash
#!/bin/bash

export CHANGED_FILES
CHANGED_FILES=$(git diff --cached --name-only --diff-filter=ACM | grep -e '\.py$')
# Format each changed file:
for file in $CHANGED_FILES; do
    black "$file" || exit 1
    git add "$file"
done

python -m unittest discover -s tests/ || exit 1

exit 0
```

you may have to install `black`

PR #943, opened 2024-09-05
Allow users to init their LoKr with perturbed normal w2

AmericanPresidentJimmyCarter:


helpers/configuration/cmd_args.py
```diff
@@ -319,6 +319,13 @@ def parse_cmdline_args(input_args=None):
         default="config/lycoris_config.json",
         help=("The location for the JSON file of the Lycoris configuration."),
     )
+    parser.add_argument(
+        "--init_lycoris_lokr_perturbed_normal",
```
bghira:
```suggestion
        "--init_lokr_norm",
```

helpers/training/trainer.py
```diff
@@ -794,6 +795,12 @@ def init_trainable_peft_adapter(self):
                     **self.lycoris_config,
                 )
 
+                if self.config.init_lycoris_lokr_perturbed_normal is not None:
+                    init_lokr_network_with_perturbed_standard(
+                        self.lycoris_wrapped_network,
+                        scale=self.config.init_lycoris_lokr_perturbed_normal,
```
bghira:
```suggestion
                        scale=self.config.init_lokr_norm,
```

PR #967, opened 2024-09-13
fix bucket worker not waiting for all queue worker to finish

burgalon:


bghira:
why? what is the background here?

burgalon:
sometimes the queue would not empty, and so a few images would drop. This became very apparent if dataset is 4 images and batch size 4, dropping 1 image would cause the training to fail since the bucket would be purged.

PR #979, opened 2024-09-16
wip: tinygrad training for flux

bghira:


bghira:
cc @sayakpaul tinygrad is so much better than pytorch for mps that i'm going to be adding it as an experimental training backend. wdyt about tinygrad support in diffusers? there is a tinygrad_flux.py file in this PR that i've pulled from the tinygrad extra distribution that includes examples for T5 and CLIP.

PR #995, opened 2024-09-26
(wip, does not work) fused bf16 optim from imoneoi

bghira:
- [ ] loads on mps
  - does not build on Apple Si yet
- [ ] loads on nvidia
  - does not have wheels for backend, does not build deps when installed via poetry
- [ ] resumes training w/o error
  - cannot test, as the optim does not load
- [ ] works with LR schedulers


bghira:
just not in a state ready for use yet.

PR #1023, opened 2024-10-03
Add the ability to use a Beta schedule to select Flux timesteps

AmericanPresidentJimmyCarter:


bghira:
should we add another option to use presets for this? similar to flux lora target, "none", "custom", and then relevant presets like "linear" or "high", "low", "middle" kinda thing?

AmericanPresidentJimmyCarter:
I would need to actually a sweep of different settings with finetunes I think. Right now I'm not entirely sure what is and isn't beneficial from a finetune perspective.

PR #1081, opened 2024-10-20
update rocm, mps and nvidia to torch 2.5

bghira:


bghira:
actually ROCm doesn't have 2.5 yet. but this fixes the source

PR #1105, opened 2024-10-27
Give --flux_schedule_auto_shift parameter priority over --flux_schedule_shift

mhirki:
I think `--flux_schedule_auto_shift` should have priority over `--flux_schedule_shift` now that the latter has a new default value after commit 4a7d42320e4457be7e52540cac1df191a1b88596. This prevents old configs from breaking assuming that people have only specified `--flux_schedule_auto_shift` in their config.


bghira:
should it actually just error out? i don't like silently selecting one or the other

bghira:
this would go into safety_check.py

mhirki:
Yeah, I guess it's better to force the user to make the choice explicitly. I'll post another pull request that just throws an error.

bghira:
thank you!

PR #1106, opened 2024-10-27
Throw an error if both --flux_schedule_auto_shift and --flux_schedule_shift are enabled.

mhirki:
As an alternative to PR #1105, just throw an error if both --flux_schedule_auto_shift and --flux_schedule_shift are enabled, forcing the user to explicitly choose one of them.

Since the default value of --flux_schedule_shift is now 3 after commit 4a7d42320e4457be7e52540cac1df191a1b88596, people need to set it 0 if they want to continue training with auto-shift.


bghira:
thanks. i'd been recommending users set one or the other for so long while doing neither by default - i would rather do auto shift by default, but the way the parameters work in python i cannot have a user pass `=true` or `=false` to update its value.

PR #1108, opened 2024-10-29
Fix gO variable name

samedii:


bghira:
ah jeeeez. thank you.

PR #1117, opened 2024-11-02
wip: omnigen initial training support attempt

bghira:
- batch size 1 only due to padding being incorrect
- high loss due to padding being incorrect
- but validations work
- LyCORIS, PEFT LoRAs work
- no controlnet style training yet


helpers/training/trainer.py
```diff
@@ -2042,6 +2071,29 @@ def model_predict(
...
+                    ),
+                    "input_image_sizes": batch.get("input_image_sizes"),
+                    "attention_mask": batch.get("encoder_attention_mask").to(
+                        self.accelerator.device
+                    ),
+                    "position_ids": batch.get("position_ids").to(
+                        self.accelerator.device
+                    ),
+                }
+                model_pred = self.transformer(**inputs)[0]
```
bghira: @staoxiao this is where i'm running the inputs for OmniGen

helpers/training/trainer.py
```diff
@@ -2228,7 +2280,26 @@ def train(self):
...
+                        u = torch.randn(bsz, device=latents.device)
+                        t = 1 / (1 + torch.exp(-u))  # t ∈ (0, 1)
+                        t = t.to(latents.device, dtype=latents.dtype)
+
+                        # Convert t to timesteps compatible with the model (scaled appropriately)
+                        timesteps = t * 999
+                        timesteps = timesteps.to(
+                            self.accelerator.device, dtype=latents.dtype
+                        )
+                        timesteps = timesteps.long()
```
bghira: @staoxiao also mostly copied your noise sampling in at this point

helpers/training/trainer.py
```diff
@@ -2344,19 +2423,25 @@ def train(self):
...
-                    )
+                    encoder_hidden_states = None
+                    if hasattr(batch["prompt_embeds"], "to"):
+                        encoder_hidden_states = batch["prompt_embeds"].to(
+                            dtype=self.config.weight_dtype,
+                            device=self.accelerator.device,
+                        )
+                        training_logger.debug(
+                            f"Encoder hidden states: {encoder_hidden_states.shape}"
+                        )
```
bghira, line 2407: this is just for compatibility with non-omnigen models

helpers/training/collate.py
```diff
@@ -528,11 +553,70 @@ def collate_fn(batch):
...
+        # input_ids = [e.get('input_ids') for e in omnigen_processed_embeddings]
+        # input_ids = torch.stack(input_ids, dim=0)
+        # TODO: Support instruction/conditioning image dropout for OmniGen.
+        # if random.random() < StateTracker.get_args().caption_dropout_probability:
+        #     instruction = '<cfg>'
+        #     latent_batch = None
+        padding_images = [e.get("padding_image") for e in omnigen_processed_embeddings]
+        extra_batch_inputs["position_ids"] = position_ids
+        extra_batch_inputs["padding_images"] = padding_images
+        extra_batch_inputs["input_ids"] = input_ids
```
bghira: @staoxiao this portion of prep work is where i ran into the most confusion and i just brute forced shapes until things worked

helpers/training/collate.py
```diff
@@ -528,11 +553,70 @@ def collate_fn(batch):
...
+            * len(latent_batch)
+        ]:
+            num_img_tokens = img_size[0] * img_size[1] // 16 // 16
+            num_text_tokens = attn_mask_len
+            total_num_tokens = num_img_tokens - num_text_tokens
+            num_tokens_for_output_images.append(total_num_tokens)
+        position_ids = OmniGenCollator.create_position(
+            attn_mask, num_tokens_for_output_images
+        )
+        # pad attn_mask to match the position_ids, eg. mask [1, 1, 1, 57] -> [1, 1, 1, 4097]
```
bghira, line 590: @staoxiao this is the padding issue i mentioned that i think is where the model's loss blows up from the start

staoxiao: I will take a look~

helpers/training/trainer.py
staoxiao: @bghira, we use a float value between 0-1 for timesteps, so there is no need to scale it `timesteps = t * 999`

helpers/training/collate.py
staoxiao: To be honest, our implementation of `atten_mask` is a little complex, integrating both unidirectional and bidirectional attention mechanisms. I suggest using our function directly: https://github.com/VectorSpaceLab/OmniGen/blob/main/OmniGen/processor.py#L241C34-L241C62

bghira: ok. it's just really not commented in there and there's no docstrings to help make use of it. we have to reverse the understanding of the code by tracing through it a lot.

```
-> x: torch.Size([1, 4, 128, 128])
-> timestep: torch.Size([1])
-> input_ids: None
-> input_img_latents: None
-> attention_mask: torch.Size([1, 4120, 4120])
-> position_ids: torch.Size([1, 4120])
The size of tensor a (4097) must match the size of tensor b (4120) at non-singleton dimension 2
```

these are the current input shapes we get from the OmniGenProcessor

Not sure why such weird shapes are coming back.

bghira: @staoxiao any ideas? currently stalled out on this because we can't get the loss to play nicely. is it a seq len issue with the length of the input embeddings?

bghira: @staoxiao if i can't figure it out i will simply be closing the pull request and not adding support, if you help it can work. otherwise it is what it is

bghira: @staoxiao ok without any upstream support it's not worth including unfortunately i will close the PR

PR #1146, opened 2024-11-12
add clip score tracking

bghira:


bghira:
image

adds new keys to wandb log tracker:

```py
        result = {
            "clip/min": min(self.eval_scores.values()),
            "clip/max": max(self.eval_scores.values()),
            "clip/mean": np.mean(list(self.eval_scores.values())),
            "clip/std": np.std(list(self.eval_scores.values())),
        }
```

bghira:
image

PR #1170, opened 2024-11-17
(experimental) Allow EMA on LoRA/Lycoris networks

bghira:
- [x] Refactor EMAModel to handle LoRA/Lycoris type networks instead of full model
- [x] Quantise the EMAModel at startup when the trained model is as well
- [x] Load state dictionary in a way that is compatible with quantised EMA model
- [x] Save state dictionary in a way that is compatible with quantised EMA model
- [x] EMA validations overhaul
  - [x] lycoris support
  - [x] peft standard LoRA
    - [x] determine a reliable method for temporarily replacing LoRA with EMA for inference
  - [x] full model checkpoint training
  - [x] optional side-by-side validations
  - [x] optional EMA-only validations
- [x] Model card details for EMA
- [x] Save/load as safetensors file instead of pytorch raw export
- [ ] Ability to export EMA model at the end of training
- [x] Export EMA model in a way that is compatible with typical adapter loading methods
  - [x] Full model export
  - [x] Standard PEFT LoRA export / loading
  - [x] Lycoris library export / loading


bghira:
extended model benchmarking to include EMA vs Checkpoint.

to demonstrate the concept this is LR 8e-3 over a single step where EMA updates every step.

image

every 100th step EMA updating at the same LR for just a single step (so EMA = base model)

image

bghira:
currently PEFT Standard LoRA is disabled because it doesn't correctly load/unload the EMA weights from the model. more development is required for that.

though now, Lycoris + EMA is working really nicely.

PR #1190, opened 2024-12-06
update sd3 seqlen to 154 max for t5

bghira:


bghira:
the 256 tokens is **total**. not just T5.

PR #1206, opened 2024-12-11
local backend: use atomicwrites library to resolve rename errors and parallel overwrites

bghira:
Removes the custom hack job logic for temp files in favour of a robust and well-used library.


bghira:
oh, also updates pytorch to 2.5.1 for nvidia systems.

PR #1211, opened 2024-12-12
fix typo

MrTuanDao:


bghira:
this isn't actually a typo. `validation_guidance_real` is the ACTUAL classifier free guidance calculation, vs `validation_guidance` which is the `flux_guidance` value.

PR #1212, opened 2024-12-12
add vae in example

MrTuanDao:


bghira:
thanks, this is a good idea to apply to all quickstarts to avoid the warning about the SDXL VAE.

PR #1276, opened 2025-01-16
Small fixes for running on AMD GPUs

rkarhila-amd:
Small modification to enable running on AMD GPUs:

1. bypass GPU mem size check if torch.cuda.version is none
2. reordered the optimizer list and added an if clause to check if bitsandbytes ademamix optimizers are available. These optimizers are not yet in the multibackend bitsandbytes for rocm.

The modified code was tested on AMD MI300X and Nvidia H100.


bghira:
thanks :wave:

PR #1294, opened 2025-01-22
add prodigy optimiser with full bf16 support

bghira:
can use `--prodigy_steps` to set the adjustment cap though it tends to stop adjusting early anyway.


bghira:
image

PR #1304, opened 2025-01-27
add stable loss calculation and tracking

bghira:
follows [this work](https://github.com/spacepxl/demystifying-sd-finetuning) to implement stable loss evaluation

this needs more testing, mostly when it's disabled or multiple eval datasets are provided.


bghira:
image

bghira:
image

bghira:
now works w/ tensorboard tracking

bghira:
fixed SD3, tested Flux, Sana, SDXL, SD1x.

fixed multiGPU handling that led to hang.

added 2nd progress bar for the total progress to be tracked.

with multiGPU though, the eval dataset is split between GPUs, but only one GPU ever runs the eval. this wastes `n-1/n` samples (n= number of GPUs). the ratio of wasted samples can be quite high with multi-node/multi-GPU cluster.

bghira:
fixed RNG save/load for CUDA devices, only the CPU RNG was getting restored before.

added `--eval_timestep_interval` which defaults to 200, which will sample every 200th timestep from the total schedule to save time while reducing accuracy.

Pevernow:
> fixed SD3, tested Flux, Sana, SDXL, SD1x.
> 
> fixed multiGPU handling that led to hang.
> 
> added 2nd progress bar for the total progress to be tracked.
> 
> with multiGPU though, the eval dataset is split between GPUs, but only one GPU ever runs the eval. this wastes samples (n= number of GPUs). the ratio of wasted samples can be quite high with multi-node/multi-GPU cluster.`n-1/n`

Perhaps we could consider running 1 or more evaluations on different validation set images on each GPU?
(I don't know the exact details, but I think it should work.)

Pevernow:
> fixed SD3, tested Flux, Sana, SDXL, SD1x.
> 
> fixed multiGPU handling that led to hang.
> 
> added 2nd progress bar for the total progress to be tracked.
> 
> with multiGPU though, the eval dataset is split between GPUs, but only one GPU ever runs the eval. this wastes samples (n= number of GPUs). the ratio of wasted samples can be quite high with multi-node/multi-GPU cluster.`n-1/n`

@bghira 
Perhaps we could consider running one or more evaluations on different validation set images on each GPU? Then take the average of the evaluation results.
(I don't know the exact details, but I think it should work.)

PR #1336, opened 2025-03-12
Add an option to completely disable validation

Slickytail:
In some cases, we want to train without doing any validation whatsoever, and without losing any time on validation (for example, in a production setting where we're going to train a bunch of loras with the same settings but with different data) or textembed caching.

At the moment, this isn't possible:
  - The training loop crashes if the validation prompt library is empty AND the unconditional prompt is disabled.
  - This means we always lose some time on precomputing the text embeddings for whatever placeholder validation prompt is present.
  - The post training validation cannot actually be disabled at all.

This PR adds a simple option that disables the attempt to precache the validation prompts, and also blocks any intermittent or post-training validation.

### DIFF
--- a/helpers/training/trainer.py
+++ b/helpers/training/trainer.py
@@ -725,7 +725,7 @@ class Trainer:
         ):
             logger.error("Cannot run validations with DeepSpeed ZeRO stage 3.")
             return
-        if self.accelerator.is_main_process:
+        if self.accelerator.is_main_process and not self.config.disable_validation:
             if self.config.model_family == "flux":
                 (
                     self.validation_prompts,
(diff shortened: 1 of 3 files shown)


bghira:
@Slickytail thanks, can you rename it `--validation_disable` to match namespacing convention

helpers/training/trainer.py
```diff
@@ -725,7 +725,7 @@ def init_validation_prompts(self):
         ):
             logger.error("Cannot run validations with DeepSpeed ZeRO stage 3.")
             return
-        if self.accelerator.is_main_process:
```
bghira, line 728: actually, all you need to do to stop validations is prevent the `self.validation` object from being created. if it is `None`, no validation occurs.

Slickytail:
I changed the name of the option.

helpers/training/trainer.py
Slickytail: If the `self.validation` object is None then no validation occurs, but the validation prompts are still encoded at the beginning of training, which loses a small amount of time. Moreover, if there are no validation prompts, then the call to `prepare_validation_prompt_list` returns None, and the tuple unpacking fails. So this change actually is necessary.

PR #1348, opened 2025-03-23
Wan 2.1 T2V training

bghira:
The 1.3B T2V model doesn't seem to work correctly in Diffusers, or, is just really bad.

I'm not sure. Some people like it, they'll train it for images. I haven't gotten any good results from it.


helpers/training/validation.py
```diff
@@ -1516,7 +1516,7 @@ def validate_prompt(
                 ]:
                     # del pipeline_kwargs["width"]
                     # del pipeline_kwargs["height"]
-                    pipeline_kwargs["num_frames"] = 1
+                    pipeline_kwargs["num_frames"] = max(81, self.args.validation_num_video_frames or 81)
```
mhirki: I think this should be changed to `max(1, self.args.validation_num_video_frames or 81)` so that 81 frames is not enforced as a minimum. I've been generating a lot of videos with only 33 frames because it's a lot faster.

bghira: sorry, i inverted the logic in my head, i'll fix that. any ideas why the images generated look so 💩 ?

bghira:
well, quanto fp8 doesn't work, but now on pytorch 2.6 and the new torchao, we have fp8-torchao working pretty well on LTXVideo and Wan 2.1.

PR #1355, opened 2025-03-27
add cfg-zero for sd3

bghira:
without
image

with
image

without (uncond)
image

with (uncond)
image


bghira:
without
image

with
image

bghira:
this PR didn't properly submit 🤦 
here's with the same prompts above

image

unconditional
image

image

PR #1364, opened 2025-03-30
Fix/keyerror latent sana model vae caching

nixczhou:
this fix is for error when training SANA model in vae caching, it will throw keyError 'latent'. Please help to review, thanks

### DIFF
--- a/helpers/caching/vae.py
+++ b/helpers/caching/vae.py
@@ -633,17 +633,26 @@ class VAECache(WebhookMixin):
 
             # Prepare final latents list by combining cached and newly computed latents
             if isinstance(latents_uncached, dict):
-                raw_latents = latents_uncached["latents"]
+                raw_latents = latents_uncached["latent"]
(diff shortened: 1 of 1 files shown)


bghira:
did you check it with LTX Video?

bghira:
i think the 'correct' fix would be:

```py
            if isinstance(latents_uncached, dict) and "latents" in latents_uncached:
```

bghira:
i ran into the same issue while adjusting for Sana support in the v2 refactor branch. i've fixed it this way [here](https://github.com/bghira/SimpleTuner/pull/1359/files#diff-5c1b47a90a9d128c083566f1d41831a1090d7fc8495ca16da5a195020d00d04b) because apparently one uses AutoencoderOutput and the other is DecoderOutput, which uses `.latent` and the other uses `.latents`

nixczhou:
thanks, then i think your refactor branch should work.

PR #1371, opened 2025-04-04
train.sh: Respect HF_HOME

clayne:
* HF_HOME was specified in the Dockerfile as /workspace/huggingface via commit 1584ddd27. When running `accelerate config`, this is the base path that will actually be used to write the config.

  Note: HF_HOME pointing to this location by default also makes sense as it provides a suitable default path from which to map volumes which may involve pre-populated models whereas mapping volumes to /root/.cache/huggingface makes a lot less sense.


clayne:
Note: opening this one against `main` as I figured your refactor branch potentially might not involve `train.sh` too much.

PR #1389, opened 2025-04-13
Auraflow training support

bghira:
Reintroduces Auraflow training support.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -29,6 +29,7 @@ model_classes = {
         "deepfloyd",
         "omnigen",
         "hidream",
+        "auraflow",
     ],
     "lora": [
         "flux",
@@ -40,6 +41,7 @@ model_classes = {
         "ltxvideo",
         "wan",
         "deepfloyd",
+        "auraflow",
     ],
     "controlnet": ["sdxl", "sd1x", "sd2x"],
 }
@@ -56,7 +58,10 @@ default_models = {
     "sana": "terminusresearch/sana-1.6b-1024px",
     "ltxvideo": "Lightricks/LTX-Video",
     "wan": "Wan-AI/Wan2.1-T2V-1.3B-Diffusers",
-    "hidream": "HiDream-ai/HiDream-I1-Full"
+    "hidream": "HiDream-ai/HiDream-I1-Full",
+    "auraflow": "terminusresearch/auraflow-v0.3",
+    "deepfloyd": "DeepFloyd/DeepFloyd-IF-I-XL-v1.0",
+    "omnigen": "Shitao/OmniGen-v1-diffusers",    
 }
 
 default_cfg = {
@@ -70,6 +75,10 @@ default_cfg = {
     "hidream": 2.5,
     "wan": 4.0,
     "sana": 3.8,
+    "omnigen": 3.2,
+    "deepfloyd": 6.0,
+    "sd2x": 7.0,
+    "sd1x": 6.0,    
 }
 
 model_labels = {
--- a/README.md
+++ b/README.md
@@ -198,6 +198,14 @@ HiDream has not been tested on 16G cards, but with aggressive quantisation and p
 
 Flux prefers being trained with multiple large GPUs but a single 16G card should be able to do it with quantisation of the transformer and text encoders.
 
+### Auraflow
+
+- A100-80G (Full tune with DeepSpeed)
+- A100-40G (LoRA, LoKr)
+- 3090 24G (LoRA, LoKr)
+- 4060 Ti 16G, 4070 Ti 16G, 3080 16G (int8, LoRA, LoKr)
+- 4070 Super 12G, 3080 10G, 3060 12GB (nf4, LoRA, LoKr)
+
 ### SDXL, 1024px
 
 - A100-80G (EMA, large batches, LoRA @ insane batch sizes)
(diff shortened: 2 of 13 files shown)


bghira:
@AstraliteHeart can you look it over maybe if you guys get a moment, i'll merge it no worries, but the high loss is a problem and the model degrades in short order:

image

it basically learns nothing, which was why AF was removed to begin with. maybe it is a problem with the Diffusers implementation under the hood that i've replicated.

but i've checked over everything, and it **inferences** fine. it's not ODE vs reverse ODE (OmniGen) and the token seq len for UMT5 around 120 or 256 or 512, none of it changes the loss either.

```
wandb:             epoch ▁▁▁▂▂▂▂▂▃▃▃▃▃▄▄▄▄▄▄▄▅▅▅▅▅▆▆▆▇▇▇▇▇▇▇▇▇▇██
wandb:       grad_absmax ▁▁▁▁▁▁▂▂▁▁▁▁▁▁▁▂▁▁▁▁▂▂▂█▄▃▅▆▃▄▆█▇▄▂▆▂▁▆▂
wandb:     learning_rate ▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁
wandb: optimization_loss ▂▃▄▂▃▃▄▂██▄▂▃▂▂▃▂▂▄▂▂▂▃▄▂▃▃▃▁▂▁▁▂▁▁▂▁▁▁▁
wandb:        train_loss ▄▆▇▇▄▄▆▆█▅▆▆▇▃▄▆█▆▄▂▃▄▄▃▄▂▂▃▃▂▃▇▂▂▂▂▁▃▃▃
wandb:   train_luminance ▂▄▆▅▃▆█▆▃▅▇▄▆▁▄▄▄▄▂▇▆▄▁▄▆▂▃▄▃▄▃▄▅▄▄▆▂▆▆▄
```

the loss starts out around 3.4-6.0 and then drops to 0.8-1.5 by the end, when the model has fully degraded and can't be used any longer for making images.

bghira:
[merged with same loss issues]

AstraliteHeart:
Thank you for bringing this back!

I've checked the loss graph and it sits just under 0.6 in the PonyFlow runs, although I've never tried training on anything less than 100k samples so perhaps a large sample is critical to keep model from collapsing?

> Other resolutions may be specified using commas to separate them: `1024x1024,1280x768,2048x2048`

Should this be updated to 1536x1536 as AF does not support 2048?

> `optimizer` - You can use any optimiser you are comfortable and familiar with, but we will use `optimi-lion` for this example.

Perhaps this is me doing something wrong but anything except AdamW had numerous issues for me in the past. Isn't lion best for large batch training, which we can't do due to VRAM limits?

> "learning_rate": "4e-5",

This looks good if we are using lion (aka 10x smaller), in my experiments 2.5e-4 is a good starting point for AdamW to get the details of the image out (although for large changes you do want something 10x larger)

> "base_model_precision": "int8-quanto",

I am not familiar with this, but does this mean we are training in 8bit? Have you checked if this may be the source of model collapse?

> The lowest VRAM OmniGen configuration is about 20-22G:

Update to AuraFlow?

> DEFAULT_LORA_TARGET = ["to_k", "to_q", "to_v", "to_out.0"]

I am still experimenting with the diffusers training code so this may be completely wrong, but shouldn't we have any of the
add_k_proj, add_q_proj, add_v_proj, to_add_out, linear_1 here?



FYI https://github.com/huggingface/diffusers/pull/10216 just got merged

bghira:
- i can update the docs to the lower resolution
- do you have a tentative HF path for ponyflow that i can add as the default flavour?
- VRAM limits are less of a problem due to the pure-bf16 optims that ST makes use of exclusively, everything supports stochastic rounding and/or kahan summation and other tricks (without this, we hit OOM pretty easily)
- the guide's using Lycoris LoKr, which needs a lower learning rate than PEFT LoRA (`--lora_type=lycoris` vs `--lora_type=standard`) and approximates more closely a full-rank tune
- the OmniGen reference was fixed on main branch afterward, i missed that initially
- for AuraFlow LoRA targets, it's probably better to mirror `--flux_lora_target` and add some specific to auraflow, but, there's not enough research/evidence for different templates yet, so i've just done the basic attention-only by default.
- our local implementation of AF does not rely on upstream classes because they're a bit slow to move on things

we've got;
- img2img support
- controlnet
- skip-layer guidance
- cfg-zero star by default

you specifically i grant license for reusing these changes in other projects and bypass AGPLv3 rules, but others still have to play ball.

configure.py
```diff
@@ -56,7 +58,10 @@
     "sana": "terminusresearch/sana-1.6b-1024px",
     "ltxvideo": "Lightricks/LTX-Video",
     "wan": "Wan-AI/Wan2.1-T2V-1.3B-Diffusers",
-    "hidream": "HiDream-ai/HiDream-I1-Full"
+    "hidream": "HiDream-ai/HiDream-I1-Full",
+    "auraflow": "terminusresearch/auraflow-v0.3",
```
hameerabbasi, line 62: Should this be the official repo? `fal/AuraFlow-v0.3`

bghira: no way. i can update and fix configs in my own repo. they're non-responsive.

PR #1413, opened 2025-04-17
Add support for JPEG-XL (file extension: jxl)

StableLlama:
Add the possibility to use images in the JPEG XL file format.

This PS doesn't change any other setting, especially it doesn't touch any places with hard coded file formats.


bghira:
why does this add the import _all over the place_?

StableLlama:
It does it everywhere `PIL` is included and then it is used to load or save an image to make sure it has this plugin.

StableLlama:
I've got git problems solving the `poetry.lock` conflict. Till it's solved it's just a draft.

bghira:
i think it's only necessary to do that in the image load function, since that's the one spot this is handled everywhere. but the default file loader is CV2, not PIL.

StableLlama:
Ok, there were only two places that are using CV2 instead of PIL to load an image. I've added a PIL fallback when CV2 doesn't understand a codec

toolkit/datasets/crop.py
```diff
@@ -105,8 +107,15 @@ def process(input_path, output_folder, detect_faces=False):
...
-                continue
+                # If OpenCV fails, try loading the image with Pillow
+                try:
+                    pil_image = Image.open(str(image_file))
+                    # Convert Pillow image to a format OpenCV can use
+                    image = cv2.cvtColor(np.array(pil_image), cv2.COLOR_RGB2BGR)
+                    return image
+                except Exception as e:
+                    logging.info(f"Image had ERROR: {input_path}")
+                    continue
```
bghira:
```suggestion
```

this is for video. it's not needed

bghira:
where you're adding the import, these scripts aren't always needing JPEG-XL. in general, i'd like to avoid having it as a baseline requirement because it sort of a rare format. we can wrap the import in a try catch and leave a note about JPEG-XL being unavailable.

StableLlama:
JPEG XL is not common right now, but it has wide momentum and is made available in many different places - and it's a great format for our usecase here. But I don't want to discuss formats, as that tends to be pointless 🙂 

Now I've made it completely optional. Right now it's by default not included, but we could easily change it to be included by default with the possibility to exclude it.

The crop.py also isn't using the CV2 to PIL fallback anymore.

PR #1430, opened 2025-04-24
hidream: remove 2x vae scaling

bghira:


bghira:
reduces hidream loss to normal range

PR #1443, opened 2025-05-11
Add support for --checkpointing_rolling_steps and --checkpointing_use_tempdir

clayne:
* --checkpointing_rolling_steps: Support creating "rolling" checkpoint saves that are independently tracked from normal checkpoint saves and which use a total limit of 1. This allows saving checkpoints more frequently in environments where spontaneous interruption is common (e.g. spot instances) but without requiring the config to specify an overly aggressive value for --checkpointing_steps.

* --checkpointing_use_tempdir: Write out all checkpoint saves to checkpoint-STEP-tmp and then atomically rename to checkpoint-STEP after success. This ensures that partially written state is never used for resuming as it's almost impossible to recover from and requires manual deletion which is a problem for automated resumes.

* init_resume_checkpoint: Move "latest" checkpoint code into checkpoint_state_latest() method which is rolling checkpoint aware.

* checkpoint_state_remove: New method to handle deleting checkpoint dirs as it's called from multiple places.

* checkpoint_state_filter: New method to determine which checkpoints should be considered candidates for keep/delete which is "suffix" aware so things like normal checkpoints vs rolling checkpoints can be handled differently. Also used with "tmp" suffixes.

* checkpoint_state_cleanup: New method to handle pruning of checkpoint dirs beyond total limit. Also used for rolling checkpoints (limit=1).

* checkpoint_state_save: New method to handle saving of checkpoints and optionally using a temporary name + rename after success when --checkpointing_use_tempdir = true.

Note: most of the new methods are reusing code that already existed.

### DIFF
--- a/helpers/configuration/cmd_args.py
+++ b/helpers/configuration/cmd_args.py
@@ -1102,10 +1102,40 @@ def get_argument_parser():
         ),
     )
     parser.add_argument(
+        "--checkpointing_rolling_steps",
+        type=int,
+        default=0,
+        help=(
+            "Functions similarly to --checkpointing_steps, but only a single rolling checkpoint is ever saved and this checkpoint does not count"
+            " towards the value of --checkpoints_total_limit. Useful for when the underlying runtime environment may be prone to spontaneous"
+            " interruption (e.g. spot instances, unreliable hardware, etc) and saving state more frequently is beneficial. This allows one to"
+            " save normal checkpoints at a reasonable cadence but save a rolling checkpoint more frequently so as to avoid losing progress."
+        ),
+    )
+    parser.add_argument(
+        "--checkpointing_use_tempdir",
+        action="store_true",
+        default=False,
+        help=(
+            "Write saved checkpoint directories to a temporary name and atomically rename after successfully writing all state. This ensures that a"
+            " given checkpoint will never be considered for resuming if it wasn't fully written out - as the state cannot be guaranteed. Useful for"
+            " when the underlying runtime environment may be prone to spontaneous interruption (e.g. spot instances, unreliable hardware, etc)."
+        ),
+    )
+    parser.add_argument(
         "--checkpoints_total_limit",
         type=int,
         default=None,
         help="Max number of checkpoints to store.",
+    )
+    parser.add_argument(
+        "--checkpoints_rolling_total_limit",
+        type=int,
+        default=1,
+        help=(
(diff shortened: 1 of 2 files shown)


bghira:
everything except atomic rename should work. in fact, we had atomic rename in the past. but it can also happen such that the instance is stopped **in the middle of that** and it still has a partial state to reference after :(

clayne:
Yeah, I saw some of the atomic renaming stuff in save_hooks, but it was for full models only - and per file oriented. This shouldn't interfere with that at all, and was just adding support to make it so the checkpoint dirs would get swapped by rename after all the save hooks finished rather than leaving any remnants around if the entire write out wasn't done (which has actually happened to me with spot style instances).

bghira:
probably --delete_bad_checkpoints would be better? it would account for filesystems where rename isn't atomic-on-crash (basically everywhere):

>  This trick doesn't work. People seem to think that this is safe becaus the POSIX spec says that rename is atomic, but that only means rename is atomic with respect to normal operation, that doesn't mean it's atomic on crash. This isn't just a theoretical problem; if we look at mainstream Linux filesystems, most have at least one mode where rename isn't atomic on crash. Rename also isn't guaranteed to execute in program order, as people sometimes expect.

from Dan Luu's FAQ from [Deconstruct 2019](https://danluu.com/deconstruct-files/)

clayne:
I'm really just trying to account for having a container wiped out in the middle of various checkpoint data files being written to the checkpoint save directory itself with regards to "atomicity." There's no way to account for all manner of things below userland anyways - like I can't predict if they'll have battery backup, writeback or writethrough cache configured, etc.. so even if rename were "atomic" at the filesystem layer we couldn't necessarily guarantee it'd absolutely hit the disk.

But to be fair, this is just about the dirent update, not a file being overwritten. It's essentially just the directory name being changed so the window is extremely small (sub-ms) vs the current window that is magnitudes larger (say 3-5 seconds for a lycoris writeout and obviously a FT way more). That's the context of "atomic" I was talking about - essentially the dirent being updated nearly instantly vs having your spot instance dropped on the floor as you're on the 3rd file of a multi-file checkpoint save.

Since these directories are essentially marked as `-tmp` it's clear they're bad. The alternative involving deleting bad checkpoints would require having knowledge of what is "good" in a given checkpoint directory, descending into the directory, ensure the sizes are non-zero etc, just way more complexy than simply using a temp dir name.

PR #1445, opened 2025-05-11
Fix missing barriers contributing to multi-process cache races

clayne:
* configure_multi_databackend: add missing is_local_main guard

* init_clear_backend_cache: add missing is_local_main guard

Both of these were contributing to multiple processes racing each other to delete the same cache files resulting in different processes having different ideas of reality depending on who won.


bghira:
can you fix the tests? it's why this hasn't been merged yet.

PR #1499, opened 2025-06-28
Update pyproject.toml for Blackwell GPU support

nitinh12:


bghira:
poetry lock should also come along with. was it not updated?

nitinh12:
> poetry lock should also come along with. was it not updated?

Done

PR #1528, opened 2025-07-02
Add support for multi-reference training (eg, with flux kontext)

Slickytail:
At the moment the training code works, and I've confirmed that the model can learn in this multi-image setup (although the output quality gets fried immediately -- probably a result of my very bad training dataset).

Validation doesn't currently work with multi-image. I'll try to get this implemented soon.

### DIFF
--- a/helpers/caching/vae.py
+++ b/helpers/caching/vae.py
@@ -48,42 +48,40 @@ def prepare_sample(
         image_path=filepath,
         model=model,
     )
-    is_cond = data_backend_id in StateTracker.get_conditioning_mappings().values()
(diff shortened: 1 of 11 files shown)


bghira:
well they didnt train dev, pro, or max with more than one input img. the frying is because the seq len is huge and out of domain just like trying to generate with odd sizes.

bghira:
@Slickytail do you want me to take this one over?

Slickytail:
Hi @bghira, I think that would be appreciated. As I mentioned, the only thing left is to make the validation work -- training works correctly.

PR #1540, opened 2025-07-06
Fix accelerator and segmentation

burgalon:


helpers/models/common.py
```diff
@@ -869,7 +869,9 @@ def uses_shared_modules(self):
         # shared modules may be when ControlNet reuses base model layers, eg. HiDream.
         return False
 
-    def get_trained_component(self, base_model: bool = False):
+    def get_trained_component(self, base_model: bool = False, unwrap_model: bool = False):
+        if not unwrap_model:
+            return self.model
```
bghira, line 874: i really don't want to change this default behaviour, that's quite a large change. why are you adding it?

burgalon: currently inductor is broken. it seems like it's completely being bypassed and not kicking in. this is fixing this

helpers/models/common.py
```diff
@@ -912,8 +914,11 @@ def _load_pipeline(
             pipeline_kwargs["watermarker"] = None
         if "watermark" in signature.parameters:
             pipeline_kwargs["watermark"] = None
-        if load_base_model:
-            pipeline_kwargs[self.MODEL_TYPE.value] = self.unwrap_model(model=self.model)
+        if load_base_model and self.model is not None:
+            if self.unwrap_model(model=self.model) is None:
+                pipeline_kwargs[self.MODEL_TYPE.value] = self.model
+            else:
+                pipeline_kwargs[self.MODEL_TYPE.value] = self.unwrap_model(model=self.model)
```
bghira, line 921: this also seems a bit round-about. what's the problem being solved?

burgalon: same as above

helpers/models/common.py
bghira: but you never call it with unwrap_model=true which is needed to get the config etc

bghira:
```
AttributeError: 'DistributedDataParallel' object has no attribute 'set_attn_processor'

```

i guess you didn't test with multi-gpu, whoops

burgalon:
will try to test the multi-gpu later this week

PR #1551, opened 2025-07-09
multi-conditioning dataset support

bghira:
continued from #1528 

for kontext, we'll add a way to switch between multiple conditions stitched-together and simply random selection between different condition sets.

for pretty much every other model it'll just randomly switch between them.

### DIFF
--- a/helpers/training/collate.py
+++ b/helpers/training/collate.py
@@ -474,13 +474,13 @@ def collate_fn(batch):
     conditioning_latents = None
 
 
+    # get multiple backend ids
+    data_backend = StateTracker.get_data_backend(data_backend_id)
     conditioning_backends = data_backend.get("conditioning_data", [])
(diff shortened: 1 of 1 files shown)


helpers/models/flux/model.py
```diff
@@ -346,7 +346,7 @@ def prepare_batch_conditions(self, batch: dict, state: dict):
             cond,
             dtype=self.config.weight_dtype,
             device=self.accelerator.device,
-            latent_channels=self.LATENT_CHANNEL_COUNT,
+            latent_channels=self.LATENT_CHANNEL_COUNT
```
bghira:
```suggestion
            latent_channels=self.LATENT_CHANNEL_COUNT,
```

bghira:
@Slickytail i think it'll work now, though there's a switch to use random vs concat mode for sampling now.

PR #1564, opened 2025-07-10
auraflow: remove double unpatchify, fix attention masking, fix LoRA target declaration

bghira:


bghira:
@astraliteheart this fixes the training issues fully.

PR #1570, opened 2025-07-13
add experimental tinygrad attention processor using tensor zero-copy

bghira:
- uses HIP backend for AMD GPUs, seems to run out of memory on 7900XTX, haven't tried larger yet (or quantisation)
- should work faster for Metal systems, compared to Torch MPS

### DIFF
--- a/helpers/models/flux/attention_tinygrad.py
+++ b/helpers/models/flux/attention_tinygrad.py
@@ -1,0 +1,496 @@
+"""
+Tinygrad-based attention processors for Flux models.
+
+This module provides drop-in replacements for standard PyTorch attention operations
+using tinygrad, with proper ROCm/AMDGPU support.
+
+Key features:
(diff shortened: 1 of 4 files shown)


helpers/models/flux/attention_tinygrad.py
```diff
@@ -0,0 +1,496 @@
...
+
+import torch
+from torch import Tensor, FloatTensor
+from tinygrad import Tensor as TinyTensor
+from tinygrad import dtypes as tinygrad_dtypes
+from einops import rearrange
+from diffusers.models.embeddings import apply_rotary_emb
+from diffusers.models.attention_processor import Attention
+import numpy as np
+
```
bghira, line 25:
```suggestion
import warnings
```

bghira:
🤔 the `.realize()` segfaults on metal

PR #1572, opened 2025-07-14
torch inductor was not working because of using unwrapped model

bghira:
cc @burgalon


burgalon:
that's awesome! thank you!

bghira:
torch inductor works on Apple MPS now as well, though i'm not sure it speeds anything up, but its doing some max-autotune GEMM stuff now at least.

PR #1575, opened 2025-07-14
kontext batched input mutation

bghira:
fixes shape error for bsz > 1 with kontext after last week's commits.


bghira:
possibly breaks the multi-image input for kontext

PR #1582, opened 2025-07-15
add scheduled huber loss option for ddpm

bghira:


helpers/configuration/cmd_args.py
```diff
@@ -122,6 +122,39 @@ def get_argument_parser():
...
     )
+    parser.add_argument(
+        "--loss_type",
+        type=str,
+        default="l2",
+        choices=["l2", "huber", "smooth_l1"],
+        help=(
+            "The loss function to use during training. 'l2' is the default, but 'huber' and 'smooth_l1' are also available."
+            " Huber loss is less sensitive to outliers than L2 loss, and smooth L1 is a combination of L1 and L2 loss."
+            " When using Huber loss, it will be scheduled via --huber_schedule and --huber_c."
```
bghira, line 133:
```suggestion
            " When using Huber loss, it will be scheduled via --huber_schedule and --huber_c."
            " NOTE: When training flow-matching models, L2 loss will always be in use."
```

PR #1599, opened 2025-07-18
Repair bug in validation stage  [update validation.py]

GuardSkill:
Repair the bug/error for multiple conditions input during the validation stage of  Kontext Lora training.

### DIFF
--- a/helpers/training/validation.py
+++ b/helpers/training/validation.py
@@ -1386,6 +1386,22 @@ class Validation:
             separator_width: Width of separator between images
             labels: Text labels for [left, middle, right] images
         """
+        # if multi condition images ,we need concat they as left image firstly
+        if isinstance(left_image, list):     
+            if all(isinstance(img, Image.Image) for img in left_image):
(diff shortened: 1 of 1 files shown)


bghira:
i need more info on the issues being solved because i dont see any errors

bghira:
so, https://github.com/bghira/SimpleTuner/pull/1575 caused this issue.

i don't think concating all the images together is the way to solve this, because they're meant to be encoded separately and then concatenated in the kontext prepare function.

GuardSkill:
> so, #1575 caused this issue.
> 
> i don't think concating all the images together is the way to solve this, because they're meant to be encoded separately and then concatenated in the kontext prepare function.

The function of `stitch_three_images` appears to be used only for DISPLAY by variable `display_validation_results `for wanDB/local; the results have been inferred by Kontext model.  This part is only for display.     If I didn't add this code ,I can't get a comparable  results in validation for multiple condition input

bghira:
oh, i thought it was editing the pipeline

PR #1620, opened 2025-07-22
save model card and full training config into checkpoint dir

bghira:


bghira:
don't need huggingface hub enabled to get a decent model card anymore.

PR #1623, opened 2025-07-23
disable SingLoRA ramp-up-steps

bghira:
harms training. without it, works fine. seems better suited for text models? or just that our LoRA init is too strong, and we can't apply weak updates.


bghira:
you know what it probably is.. the grad_absmax is 0 for the first few steps and in general i think it's vanishing a bunch of gradients even once the absmax is non-zero.

they use fp16 in the original examples for the precision, but we don't have the issue with bf16 without ramp-up-up steps because we don't cause such small gradients.

Issue #1611, opened 2025-07-21
Consider adding TREAD training


AmericanPresidentJimmyCarter:
https://github.com/CompVis/tread

It sounds more complicated than it is.

You start the whole training with all the tokens, then at an early block in the transformer you make your batch tensor's sequence a random subset of tokens from that batch. Then you put these random tokens < total tokens through lots of layers, then at a later block or the last block you add the tokens back in so you backprop on the entire image.

#### Pick your random tokens (one per "route")

```py
if use_routing and idx == self.routes[route_count]['start_layer_idx']:
    x_D_last = x.clone()
    mask_info = self.router.get_mask(
        x,
        mask_ratio=self.routes[route_count]['selection_ratio']
    )
    x = self.router.start_route(x, mask_info)
```

#### Run it through your block(s)

```py
# Still in that same loop:
x = block(x, c)
```

#### Reassemble

```py
if use_routing and idx == self.routes[route_count]['end_layer_idx']:
    x = self.router.end_route(
        x,             # the shortened tensor (B, num_keep, D)
        mask_info,
        original_x=x_D_last
    )
    fp32_next = True
    if route_count < len(self.routes) - 1:
        route_count += 1
```

Altogether, it's like:

```py
routes=[
    {'selection_ratio': 0.5, 'start_layer_idx': 3, 'end_layer_idx': 6},
    {'selection_ratio': 0.3, 'start_layer_idx': 7, 'end_layer_idx': 10}
]

use_routing = self.training and self.enable_routing and self.routes
route_count = 0 if use_routing else None

for idx, block in enumerate(self.blocks):
    # ==== START of Route route_count ====
    if use_routing and idx == self.routes[route_count]['start_layer_idx']:
        # sample one mask for this route
        mask_info = self.router.get_mask(x, mask_ratio=self.routes[route_count]['selection_ratio'])
        x_D_last  = x.clone()                   # save full tokens
        x         = self.router.start_route(x, mask_info)  # truncate to kept tokens

    # process the (possibly truncated) x through block
    x = block(x, c)

    # ==== END of Route route_count ====
    if use_routing and idx == self.routes[route_count]['end_layer_idx']:
        # splice back the skipped tokens for this route
        x, fp32_next = self.router.end_route(x, mask_info, original_x=x_D_last), True
        # move on to the next route (if any)
        if route_count < len(self.routes) - 1:
            route_count += 1
```

They claim about a 40x training speedup and saving on compute and memory.

PR #1672, opened 2025-07-27
Add TREAD training for the FLUX model

AmericanPresidentJimmyCarter:
Closes https://github.com/bghira/SimpleTuner/issues/1611

I haven't had time to test it much, but you can make route configs (CLI args) like:

```
            f'--tread_config={json.dumps({
                'routes': [
                    {
                        'start_layer_idx': 8,
                        'end_layer_idx': -8,
                        'selection_ratio': 0.5,
                    },
                ],
            })}',
```

It works, it does speed up training, and the speedup is proportion to the `selection_ratio` (the closer that is to 1.0, the more tokens are dropped).

I haven't messed with the router configurations much to see what works and what doesn't.

It also worked with masked loss, where masking prevents certain tokens from being dropped. This slows down training by using more tokens.

The token dropping was implemented for RoPE in FLUX, as this is required to make TREAD work. I'm unsure the implementation is 100% correct, but it seems to work.

The more tokens you drop, the higher loss seems to be when you start LoRA/LoKr training. It tends to correct fairly rapidly and the images look more normal. I'm unsure if it's the network just adjusting to the smaller amount of tokens being supplied in the intermediary layers.

There was also a bug that seemed to break training with masked loss for flux in the main branch. This was fixed with a `self.config.model_flavour == "kontext"` guard.


bghira:
replaced by #1675

PR #1680, opened 2025-07-29
add Wan TREAD support and docs

bghira:
really fun speedup to experiment with.
### No TREAD
https://github.com/user-attachments/assets/77f7fac7-ddfd-4bcc-ac0a-d03a36bfff1b

### U-shaped TREAD config (multiple routes)
https://github.com/user-attachments/assets/5ffa9362-df89-496c-92f0-4e7449efcec5

### Simple 0.1 TREAD config
https://github.com/user-attachments/assets/240f5598-a77a-4325-89ce-54df656fd905

### DIFF
--- a/config/examples/multidatabackend-small-video-480p.json
+++ b/config/examples/multidatabackend-small-video-480p.json
@@ -1,0 +1,28 @@
+[
+  {
+    "id": "dreambooth-480",
+    "type": "huggingface",
+    "dataset_type": "video",
+    "dataset_name": "sayakpaul/video-dataset-disney-organized",
+    "crop": false, "crop_style": "random", "crop_aspect": "square",
+    "minimum_image_size": 128,
+    "maximum_image_size": 480,
+    "target_downsample_size": 480,
+    "resolution": 480,
+    "resolution_type": "pixel_area",
+    "metadata_backend": "huggingface",
+    "caption_strategy": "huggingface",
+    "huggingface": {
+      "video_column": "video",
+      "caption_column": "prompt"
+    },
+    "cache_dir_vae": "cache/vae/{model_family}/dreambooth-480"
+  },
+  {
+    "id": "alt-embed-cache",
+    "dataset_type": "text_embeds",
+    "default": true,
+    "type": "local",
+    "cache_dir": "cache/text/{model_family}"
+  }
+]


bghira:
@stefan-baumann if you get a chance, i'm curious if you see problems here with this video model approach for TREAD.

stefan-baumann:
Hi, thanks for tagging me on this, this looks really interesting!

Also awesome to see that you integrated our work!

 I'm currently on vacation for the rest of the week, so I cannot give detailed feedback beyond my opinion that TREAD should, in general, also be applicable to video generation. I'll ping Felix Krause about this in hopes that he can give you better feedback :)

flixmk:
Hey, thanks for trying out TREAD! 
Our general conclusion was that this routing also works without the additional masking loss which makes it a bit less hyperparameter-heavy. 
We have not yet extensively tried TREAD on video. I can only offer my intuition which tells me that TREAD should work on video. However, one thing to note is that different modalities might benefit from different routing setups.

bghira:
i was using a config like this for the 29 layer Wan 1.3B model:

```json
{
  "tread_config": {
      "routes": [
          { "selection_ratio": 0.1, "start_layer_idx": 2, "end_layer_idx": 8 },
          { "selection_ratio": 0.25, "start_layer_idx": 9, "end_layer_idx": 11 },
          { "selection_ratio": 0.35, "start_layer_idx": 12, "end_layer_idx": 15 },
          { "selection_ratio": 0.25, "start_layer_idx": 16, "end_layer_idx": 23 },
          { "selection_ratio": 0.1, "start_layer_idx": 24, "end_layer_idx": -2 }
      ]
  }
}
```

but i've not implemented any ramp up scheduling. was there experimentation done with scheduling the token dropping?

flixmk:
We have not found a schedule that worked better than what we propose in our arxive ([https://arxiv.org/pdf/2501.04765](https://arxiv.org/pdf/2501.04765)). But we only looked at training from scratch so far and not at finetuning existing models so it might make a lot of sense to apply some scheduling here. 
We also explored multiple consecutive routes which does give a slight edge. However, its a bit more hyperparameter prone and since route location (start / end) can be quite important, I would suggest following the setting shown in the paper (single route from 2->L-4; L being the depth of the transformer) and using that as a starting point. This scaled quite well accross model sizes, resolutions and tasks that we explored.

bghira:
cc @AmericanPresidentJimmyCarter 

we used 2, L, -2. but i'll give it a try to exclude more end layers. and i had the same thought that training from scratch likely aligns perfectly with the objective, but finetuning probably needs to ease into it.

AmericanPresidentJimmyCarter:
From my testing of Flux and SD3.5, usually the first 6-8 layers were critical and then it was similar at the end. In the Wan dense models, it seemed like the middle blocks did quite a bit more and overall the models were trained to a level where they became more important.

Since TREAD doesn't fully skip blocks, you might be able to get away with more tokens dropped closer to the beginning and ends.

You might want to try something more aggresive like:

```py
{
  "tread_config": {
      "routes": [
          { "selection_ratio": 0.25, "start_layer_idx": 4, "end_layer_idx": 8 },
          { "selection_ratio": 0.35, "start_layer_idx": 9, "end_layer_idx": 11 },
          { "selection_ratio": 0.5, "start_layer_idx": 12, "end_layer_idx": 27 },
          { "selection_ratio": 0.35, "start_layer_idx": 28, "end_layer_idx": 32 },
          { "selection_ratio": 0.25, "start_layer_idx": 33, "end_layer_idx": 36 }
      ]
  }
}
```

For Wan 14b.

PR #1691, opened 2025-08-03
remove bad iCCP chunk from PNG data before passing bytes to CV2 (silence libpng console spam)

bghira:
suppresses spam from libpng by removing the bad data


helpers/image_manipulation/load.py
```diff
@@ -1,10 +1,11 @@
 import logging
 import tempfile
-import os
+import os, sys
```
bghira:
```suggestion
import os
```

helpers/image_manipulation/load.py
```diff
@@ -5,6 +5,7 @@
 
 from io import BytesIO
 from typing import Union, IO, Any
+from contextlib import contextmanager
```
bghira:
```suggestion
```

PR #1701, opened 2025-08-06
VAECache needs to shuffle bucket contents before splitting across GPUs

bghira:
when some cache is preserved across restarts, work units are distributed unevenly across GPUs, as, one or more GPUs may contain solely corrupted samples.

the reduction in bucket contents went from 79k to 4 samples, but when shuffling before distribution, we'd get better results and each worker receives 19k remaining samples to process.

removed `do_shuffle` parameter for reduce_buckets because it's been split by this point and shuffling does nothing.

### DIFF
--- a/helpers/caching/vae.py
+++ b/helpers/caching/vae.py
@@ -1312,7 +1312,7 @@ class VAECache(WebhookMixin):
                             current=statistics["total"],
                         )
                     self.debug_log(
-                        f"Completed process_buckets, all {len(futures)} futures have been returned."
+                        "Completed process_buckets, all futures have been returned."
                     )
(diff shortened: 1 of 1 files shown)


helpers/caching/vae.py
```diff
@@ -1309,7 +1312,7 @@ def process_buckets(self):
                             current=statistics["total"],
                         )
                     self.debug_log(
-                        "Completed process_buckets, all futures have been returned."
+                        f"Completed process_buckets, all {len(futures)} futures have been returned."
```
bghira:
```suggestion
                        "Completed process_buckets, all futures have been returned."
```

PR #1703, opened 2025-08-07
Fix crash in Cosmos2 latent generation

AmericanPresidentJimmyCarter:


bghira:
did i accidentally re-push this one? i was also working on the VAE cache logic, but i tried to keep this change in-tact. sorry about that.

PR #1722, opened 2025-08-20
Bugfix: wan vae normalisation + deterministic posterior sampling

squewel:
This fixes:

1. Fixes vae input normalization from [0,1] (no normalization) to [-1,1] range. Lack of normalization resulted in overexposed/washed-out inputs
2. Posterior sampling switched to .mode(), which greatly improves input quality. With .sample() -- previous behavior -- training outcomes were artifacted


squewel:
I'm not sure what you want to do with posterior computation so I just put .mode() for now

PR #1767, opened 2025-10-17
Chroma HD

bghira:


bghira:
working on implementing ControlNet LoRA and then this will be merged.

PR #1769, opened 2025-10-18
Wan 2.2 A14B, group offload memory optimisations

bghira:


bghira:
wan 2.2 ti2v 5b works great now. 2.2 t2i and 2.2 i2v a14b also work well.

wan 2.1 i2v 14b is likely going to be weird, it has CLIP embeds i've not yet quite figured out.

recommended to train 2.2 ti2v 5b or 2.2 t2v/i2v.

group offload was working with layer strategy after adjusting to exclude text encoders and vae from the offload policy.

PR #1829, opened 2025-10-25
ponyflow: add path and override sequence length to 768

bghira:


aminmoghtader:
Follow and connect for programmers to connect and collaborate. We programmers need to have a global connection with each other
Thank you very much

PR #1866, opened 2025-10-29
Add non-Python files to SimpleTuner wheel.

hameerabbasi:
See https://setuptools.pypa.io/en/latest/userguide/miscellaneous.html#controlling-files-in-the-distribution

By default, only `*.py` files are included in the distribution when using setuptools. This means that other files, such as HTML files, are missing.

To fix this issue, we include them in `MANIFEST.in`.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+recursive-include simpletuner *.html *.css *.js *.ico *.json *.md


hameerabbasi:
Currently, `SimpleTuner` installed with pip errors out on missing HTML files and gives a 500 error code.

PR #1886, opened 2025-10-31
FSDP2: Fix save/load, introduce CPU offload options, fix gradient checkpointing

bghira:
it wasn't properly selecting blocks to shard, and two FSDP options weren't exposed to the user, cpu offload and activation checkpointing.

we can't use torch's activation checkpointing with FSDP, we have to use FSDP-aware version. i think maybe this is FSDP2 limitation, perhaps - but it's faster to use the correct implementation anyway.

### DIFF
--- a/simpletuner/__init__.py
+++ b/simpletuner/__init__.py
@@ -70,6 +70,8 @@ def _suppress_swigvarlink(message, *args, **kwargs):
         return None
     if "MPS autocast" in text:
         return None
+    if "remove second argument of ws_handler" in text:
+        return None
 
     return _original_warn(message, *args, **kwargs)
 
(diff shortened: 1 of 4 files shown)


bghira:
we can train Flux on A5000 with 24G, all modules and layers.

same for full-rank training of SD3.5 Large, and medium.

simpletuner/helpers/training/diffusers_overrides.py
```diff
@@ -375,3 +380,65 @@ def enable_reversible_fusion():
...
+                    dist.broadcast(shard, src=0, group=dist.group.WORLD)
+                shards[name] = shard
+
+        model.load_state_dict(shards, assign=True)
+        return model
+
+    accelerate_fsdp_utils.fsdp2_load_full_state_dict = replacement
+
+
+print("Override FSDP2 state dict loader")
```
bghira:
```suggestion
```

PR #1887, opened 2025-11-01
Fix sd3.5 large link 

dill-shower:


bghira:
is it still used anywhere? if it is, is possible that simply removing it will be ok, since these are defined in the SD3 class now.

dill-shower:
> is it still used anywhere? if it is, is possible that simply removing it will be ok, since these are defined in the SD3 class now.

Yes. On my config I need fix this manually on file. SimpleTuner trying download files from this link and get error

PR #1921, opened 2025-11-06
Fix AttributeError: 'Namespace' object has no attribute 'cache_file_suffix'

ZiyiTsang:
**Problem**: Direct access to `self.args.cache_file_suffix` at line 1668 in `simpletuner/simpletuner/helpers/data_backend/factory.py` caused AttributeError when users didn't specify the optional `--cache_file_suffix` parameter.

**Solution**: Changed `cache_file_suffix=self.args.cache_file_suffix` to `cache_file_suffix=getattr(self.args, "cache_file_suffix", None)` to provide safe default handling. This follows the established pattern in the codebase for optional arguments and maintains backward compatibility.

**Files Changed**: `simpletuner/simpletuner/helpers/data_backend/factory.py` (1 line)

### DIFF
--- a/simpletuner/helpers/data_backend/factory.py
+++ b/simpletuner/helpers/data_backend/factory.py
@@ -1665,7 +1665,7 @@ class FactoryRegistry:
                 accelerator=self.accelerator,
                 instance_data_dir=init_backend.get("instance_data_dir", ""),
                 cache_dir=init_backend["cache_dir"],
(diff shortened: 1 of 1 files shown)


bghira:
the deeper issue is that VAECache shouldn't be instantiated for image_embeds backends, i'm working on a more comprehensive fix, but if this works for you, you should be able to continue training.

ZiyiTsang:
ok

PR #1929, opened 2025-11-07
add --validation_preview option to ship intermediary samples to the webhook

bghira:
shows validation outputs on every step.

is work-in-progress, to be expanded with option to limit the frequency of generating the int validation samples.

should work for videos too.


bghira:
using the preview option brings up a modal like this.

https://github.com/user-attachments/assets/a04b2971-277a-4133-9f8a-ed2245ddac96

PR #2088, opened 2025-12-06
align computation strip() and retrieval which does not strip()

bghira:
This pull request enhances the handling and validation of the `instanceprompt` caption strategy in the `magic_prompt` function. It ensures that prompts are properly cleaned (trimmed of whitespace and empty prompts removed), and adds error handling for missing or empty prompts. Corresponding unit tests have been added to verify this behavior.

**Improvements to prompt handling and validation:**

* Improved `magic_prompt` in `prompts.py` to trim whitespace from `instance_prompt` values, remove empty prompts from lists, and raise clear errors if no valid prompt is provided.

**Testing enhancements:**

* Added unit tests in `test_prompthandler.py` to verify that single and list `instance_prompt` values are correctly cleaned and validated.

### DIFF
--- a/simpletuner/helpers/prompts.py
+++ b/simpletuner/helpers/prompts.py
@@ -374,7 +374,19 @@ class PromptHandler:
                 sampler_backend_id=sampler_backend_id,
             )
         elif caption_strategy == "instanceprompt":
-            return instance_prompt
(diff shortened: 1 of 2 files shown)


bghira:
this resolves condition where text embed cache entry cannot be found.

PR #2213, opened 2025-12-22
fix: remove unsupported hidden_states_buffer from SDXL model_predict

joeqzzuo:
UNet2DConditionModel from diffusers does not support the hidden_states_buffer
  parameter, unlike custom transformers (Flux, AuraFlow, LTX Video, etc.).

  This caused training to fail with:
  TypeError: UNet2DConditionModel.forward() got an unexpected keyword argument 'hidden_states_buffer'

  Tested with diffusers 0.36.0 and simpletuner 3.3.1.


bghira:
thanks, forgot to remove that after experimenting with it.

for reference, i'd flattened the hidden states from matching down/up blocks to treat them as image tokens for comparison purposes, but it doesn't seem to provide any benefit due to the architecture's skip connections that already effectively do the same thing.

PR #2225, opened 2025-12-23
sdnq quant engine

bghira:
This pull request adds support for SDNQ (SD.Next Quantization), a cross-platform quantization engine optimized for training on AMD, Apple, and NVIDIA hardware. SDNQ enables significant memory savings and efficient training through stochastic rounding and quantized optimizer states. The update introduces new SDNQ acceleration presets across all major model classes, updates documentation, and adds `sdnq` as a dependency.

**SDNQ Integration and Presets:**

- Added SDNQ as a new acceleration backend (`SDNQ`) in `AccelerationBackend`, and implemented a utility function `get_sdnq_presets` to generate standard SDNQ acceleration presets for various quantization levels. These presets are now included in the acceleration options for all major model classes, providing users with cross-platform memory optimization choices. [[1]](diffhunk://#diff-34d0a1984c0a965c5b6c376398ce7faea14efbddbd40435846f4052c9445c953R22) [[2]](diffhunk://#diff-29334f367fcdb0a02fcd7f4033d00bf842e24d8142de41fd161193c85fd3fafeR28-R107) [[3]](diffhunk://#diff-e34e7b7e69a677399426b34f91fd1a10bb8e9f9f5a9eea54c8de088c103acebbL23-R23) [[4]](diffhunk://#diff-e34e7b7e69a677399426b34f91fd1a10bb8e9f9f5a9eea54c8de088c103acebbR219-R220) [[5]](diffhunk://#diff-bb687de4a71d62f31bb7b892a22f0a94108d93d0323112a0bc221680a2890b5fL9-R9) [[6]](diffhunk://#diff-bb687de4a71d62f31bb7b892a22f0a94108d93d0323112a0bc221680a2890b5fR209-R210) [[7]](diffhunk://#diff-571111f5431ee4a79012ae61ee50c3b6dffca3fbe3ccbdeba0a3132ca48343e7L10-R10) [[8]](diffhunk://#diff-571111f5431ee4a79012ae61ee50c3b6dffca3fbe3ccbdeba0a3132ca48343e7R198-R199) [[9]](diffhunk://#diff-01ff9e59af344fe6857ccb05e1e335ed6c7029d571dfae4d6014dea582b36cfeL11-R11) [[10]](diffhunk://#diff-01ff9e59af344fe6857ccb05e1e335ed6c7029d571dfae4d6014dea582b36cfeR198-R199) [[11]](diffhunk://#diff-68d87fba290a2ecb1f23edb2f5c8b5f44f32ad2726ad772612d1450947b81845L12-R12) [[12]](diffhunk://#diff-68d87fba290a2ecb1f23edb2f5c8b5f44f32ad2726ad772612d1450947b81845R260-R261) [[13]](diffhunk://#diff-5aa130951df78ab03d60ca20b1126a94d40fb0035ac90907eb33b3de05345489L12-R12) [[14]](diffhunk://#diff-5aa130951df78ab03d60ca20b1126a94d40fb0035ac90907eb33b3de05345489R227-R228) [[15]](diffhunk://#diff-2390d666883fb499d80e71fbe6b05b21e5141f934149c7fc66937368b6ae0c02L17-R17) [[16]](diffhunk://#diff-2390d666883fb499d80e71fbe6b05b21e5141f934149c7fc66937368b6ae0c02R249-R250) [[17]](diffhunk://#diff-df6046a31196ac9ebae68e69c519f1db8029fb3d44f975605572c7d42401211dL10-R10) [[18]](diffhunk://#diff-df6046a31196ac9ebae68e69c519f1db8029fb3d44f975605572c7d42401211dR234-R235) [[19]](diffhunk://#diff-7415b0e19fb3d9edd0f57963ccdbfc2ea8946e23297f845a1f4948b75bead27bL20-R20) [[20]](diffhunk://#diff-7415b0e19fb3d9edd0f57963ccdbfc2ea8946e23297f845a1f4948b75bead27bR214-R215) [[21]](diffhunk://#diff-0e4e8215bf58377859f3a1fdcdc6830fc78a66180571a8644da1d0fc68214e25L10-R10) [[22]](diffhunk://#diff-0e4e8215bf58377859f3a1fdcdc6830fc78a66180571a8644da1d0fc68214e25R205-R206) [[23]](diffhunk://#diff-fc3d33039ac332c8aaaec578562f232b62a1db1b0cae4847e8d3e94be779bbeaL4-R6)

**Dependency and Setup Updates:**

- Added `sdnq>=0.1.2` to the list of required dependencies in `setup.py` to ensure SDNQ is available for use in all supported environments.

**Documentation Enhancements:**

- Expanded the documentation to include a comprehensive section on SDNQ, detailing its features, recommended precision levels, optimizer options, and practical usage notes for training and inference.

### DIFF
--- a/simpletuner/configure.py
+++ b/simpletuner/configure.py
@@ -670,6 +670,15 @@ class MemoryPresetsSession:
                 # Clear custom block swap when selecting RamTorch or Group Offload
                 if backend_name in {"RAMTORCH", "GROUP_OFFLOAD"}:
                     self.custom_block_swap_count = 0
+            # If preset has a group, deselect other presets in the same group
+            if preset.group:
+                for other_preset in self.presets:
+                    if (
+                        other_preset.group == preset.group
+                        and other_preset.backend.name != backend_name
+                        and other_preset.backend.name in self.selected_presets
+                    ):
+                        del self.selected_presets[other_preset.backend.name]
             # Select this preset
             self.selected_presets[backend_name] = preset.level
 
--- a/simpletuner/helpers/acceleration/__init__.py
+++ b/simpletuner/helpers/acceleration/__init__.py
@@ -1,6 +1,21 @@
 """Acceleration backend infrastructure for memory optimization presets."""
 
 from .backends import AccelerationBackend
-from .preset import AccelerationPreset, get_sdnq_presets
+from .preset import (
+    AccelerationPreset,
+    get_bitsandbytes_presets,
+    get_deepspeed_presets,
+    get_quanto_presets,
+    get_sdnq_presets,
+    get_torchao_presets,
+)
 
-__all__ = ["AccelerationBackend", "AccelerationPreset", "get_sdnq_presets"]
+__all__ = [
+    "AccelerationBackend",
+    "AccelerationPreset",
+    "get_bitsandbytes_presets",
+    "get_deepspeed_presets",
+    "get_quanto_presets",
+    "get_sdnq_presets",
+    "get_torchao_presets",
+]
--- a/simpletuner/helpers/acceleration/backends.py
+++ b/simpletuner/helpers/acceleration/backends.py
@@ -20,3 +20,6 @@ class AccelerationBackend(Enum):
     DEEPSPEED_ZERO_3 = auto()
     FSDP2 = auto()
     SDNQ = auto()  # SD.Next Quantization - works on AMD, Apple, NVIDIA
+    TORCHAO = auto()  # TorchAO quantization - NVIDIA only
+    QUANTO = auto()  # Quanto quantization - works on AMD, Apple, NVIDIA
+    BITSANDBYTES = auto()  # BitsAndBytes quantization - NVIDIA only
--- a/simpletuner/helpers/acceleration/preset.py
+++ b/simpletuner/helpers/acceleration/preset.py
@@ -24,6 +24,9 @@ class AccelerationPreset:
     tradeoff_notes: str  # Requirements, warnings
     requires_cuda: bool = False
     requires_min_system_ram_gb: int = 0
+    requires_model_type: list[str] | None = None  # If set, only show for these model types (e.g. ["full"])
+    group: str | None = None  # Mutually exclusive group (e.g. "quantization" - only one can be selected)
+    display_group: str | None = None  # UI grouping (e.g. "deepspeed" to show ZeRO 1/2/3 as one card)
     config: dict[str, Any] = field(default_factory=dict)  # Config keys this preset sets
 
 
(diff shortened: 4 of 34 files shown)


Disty0:
SDNQ supports low bits from 1 to 7 too.  
I do not recommend below 8 bits for full finetune but they can be used for lora training.  

Relevant types i recommend are:  
 - uint8, uint16 or float16 (fp16 in sdnq is quantized too) for full finetuning or on the optimizers.  
 - int8, int6, int5 or uint5, uint4, uint3 (maybe uint2 too if you really need it) if you are not going to update the model weights (aka only for lora training).  
 - fp8 types are also available but i don't really recommend them.  
 - int7 is also available but not recommended because it is slow and not that much smaller than int8.  

I would also highly recommend using SVD at 8 svd_steps for anything below 5 bits.  
But i won't recommend using SVD if you are going to full finetune / actively update the model weights as re-calculating SVD is expensive.  

I do upload some uint4 svd prequants here: https://huggingface.co/collections/Disty0/sdnq  
You need to convert them to sdnq training with convert_sdnq_model_to_training after loading the model normally.  
SDNQ also needs to be imported before loading them as sdnq needs to be registered into Diffusers first.  

You can use SVD in post load quant and SDNQConfig too, but SVD takes a little bit longer to quantize and it is not deterministic.  
These 2 reasons are why i upload SDNQ SVD quants to HF.  

16 bit quantization is also available if you want the maximum precision on full finetuning (aka Stable Cascade).  
int16, uint16 and float16 (quantized) are availabe. I would recommend uint16 from the 16 bit types.

## quantization_device:  
quantization_device and return_device are intended for using the GPU for quantization without going OOM.  
You can load the model into "cpu" and set quant device to "cuda" and return device to "cpu" for much faster quantization without OOM.  
Leaving them empty / None won't touch the model devices.  

## modules_to_not_convert:  
I also maintain a list in sdnq too: https://github.com/Disty0/sdnq/blob/main/src/sdnq/common.py#L134-L220  
Also i would recommend using ".proj_out" instead of "proj_out" for flux as it has inner layers named "proj_out" too.  
A key starting with "." means the root level on SDNQ.  

## quantize_activations / use_quantized_matmul:  
Matmul precision and weights are actually decoupled. You can use INT8 / FP8 / FP16 matmul with any weight dtype or even no quantized weights.  
Nvidia consumer GPUs or anything non Nvidia should prefer INT8 matmul over FP8.  
Nvidia datacenter GPUs can be faster with FP8 matmul.  
FP16 (Quantized) matmul is also available if you want the maximum precision (again, Stable Cascade).  

## Muon:  
Muon also has use_quantized_matmul, quantized_matmul_dtype and zeropower_dtype options.  
zeropower_dtype comes from Muon base and sets matmul precision in zeropower_via_newtonschulz5.  
Using use_quantized_matmul ignores zeropower_dtype and will run that matmul in int8 / fp8 / fp16 (whatever you set in quantized_matmul_dtype).  

Muon also has "adamw_*" and "muon_*" prefixes on args too. You can use them to set different args for AdamW fallback and Muon.  
For example, "muon_use_quantized_buffers=True" and "adamw_use_quantized_buffers=False" will only set "use_quantized_buffers" for Muon and will disable it for AdamW.  
Tho you don't need this if you are doing proper param grouping and passing the "use_muon" arg per group.  


## Saved model weights:  
I only save the quantized model weights and use the dequantize_sdnq_training.py script to dequantize manually if i want to use that model: https://github.com/Disty0/sdnq/blob/main/scripts/dequantize_sdnq_training.py
This is more of a personal preference as i am short on disk space.

PR #2422, opened 2026-01-16
add validation_guidance_real to Flux2 pipeline

bghira:
This pull request introduces support for real classifier-free guidance (CFG) in the FLUX.2 model pipeline, refines how negative prompts are handled, and improves validation logic to accommodate these changes. The updates ensure correct handling of negative embeddings and configuration parameters for both training and inference.

### Real CFG support and negative prompt handling

* Added `VALIDATION_USES_NEGATIVE_PROMPT = True` to the `Flux2` model class to signal that negative prompts are required for real CFG.
* Updated the `convert_negative_text_embed_for_pipeline` method to properly check for real CFG requirements, handle `negative_text_ids`, and only return negative embeddings when needed.
* The pipeline's `__call__` method now accepts new arguments: `guidance_scale_real`, `no_cfg_until_timestep`, `negative_prompt_embeds`, and `negative_text_ids`, enabling real CFG during inference.
* Added logic in the pipeline to validate that negative embeddings are provided when real CFG is requested, raising an error otherwise.
* Implemented a second inference pass for real CFG: when `guidance_scale_real > 1.0`, the pipeline computes noise predictions using negative embeddings and blends them with the positive prediction as required for classifier-free guidance.

### Validation logic improvements

* Updated validation to pass `no_cfg_until_timestep` for both `flux` and `flux2` model families, ensuring correct behavior during validation runs.

### DIFF
--- a/simpletuner/helpers/training/save_hooks.py
+++ b/simpletuner/helpers/training/save_hooks.py
@@ -199,31 +199,58 @@ class SaveHookManager:
 
     def _build_tag_frequency_from_captions(self) -> dict[str, dict[str, int]]:
         """
-        Scan all captions from text embed caches, split into tags, and build frequency dict.
+        Scan all training captions, split into tags, and build frequency dict.
         Returns dict of dataset_id -> {tag: count}.
         """
         import re
 
+        from simpletuner.helpers.prompts import PromptHandler
+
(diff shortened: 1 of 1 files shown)


bghira:
before, after
image

PR #2621, opened 2026-02-13
s2v/ltx-2 audio auto split should be enabled more intelligently

bghira:
This pull request introduces several improvements and bug fixes related to audio data handling, error tracking, and configuration logic in the `simpletuner` project. The most notable changes enhance error reporting for audio processing, refine the logic for auto-injecting audio configurations, and expand test coverage for these behaviors.

**Audio Processing and Error Reporting Improvements:**

* Audio processing errors are now explicitly tracked in the `statistics` dictionary under the `processing_error` key, ensuring failures are no longer silently ignored. This enables better diagnostics and downstream reporting.
* The `BucketReport` class now provides more actionable recommendations, including suggestions when videos lack audio streams or when audio processing fails.

**Audio Configuration Injection Logic:**

* The logic for auto-injecting audio configs in `_inject_s2v_audio_configs` is now more precise: audio configs are only auto-injected when the model requires S2V datasets or when the user has explicitly opted in by providing an audio config section. This prevents unwanted audio dataset creation for models that merely support (but do not require) audio.

**Test Coverage Enhancements:**

* New and expanded tests verify that:
  - Audio processing errors are tracked in statistics.
  - `BucketReport` generates appropriate recommendations for missing audio or processing errors.
  - Audio config injection respects whether the model requires or merely supports audio, and whether the user has provided an explicit audio config.

**Bucket Report Statistics Update:**

* The `compute_aspect_ratio_bucket_indices` method now updates the bucket report with filtering statistics if available, improving the accuracy of post-processing reports.

### DIFF
--- a/documentation/DATALOADER.es.md
+++ b/documentation/DATALOADER.es.md
@@ -125,6 +125,42 @@ Los backends de audio admiten un bloque `audio` dedicado para que los metadatos 
 - **`target_resolution`** – resolución de video objetivo para el modo solo audio (usada para calcular dimensiones de latentes).
 - La configuración estándar de audio (número de canales, directorio de caché) se mapea directamente al backend de audio en tiempo de ejecución creado por `simpletuner.helpers.data_backend.factory`. Se evita el padding intencionalmente: los clips se truncan en lugar de extenderse para mantener el comportamiento consistente con entrenadores de difusión como ACE-Step.
 
+#### Configuración de audio para entrenamiento S2V
+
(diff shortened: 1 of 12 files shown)


bghira:
@kaibioinfo this is for your "no samples detected" error with silent audio samples. the default (where `audio = {}`) means that `auto_split` defaults to `True` and the auto-generated dataset is created with `source_from_video=True` and `allow_zero_audio=False`. this caused all samples to be rejected, because zero samples were.

the fix in the dataloader configuration factory is to stop auto-creating the audio split. it's a bit unfortunate, i'd rather the UX for users with audio data be better (it should be detected automatically) but not worth trying to chase the problems caused by it being enabled by default.

PR #2677, opened 2026-04-10
Security: Webhook URL accepted without SSRF protections

tomaioo:
## Summary

Security: Webhook URL accepted without SSRF protections

## Problem

**Severity**: `Medium` | **File**: `simpletuner/helpers/webhooks/config.py:L6`

Webhook configuration only checks for field presence and allows arbitrary `webhook_url`/`callback_url` values. If non-admin or untrusted users can set webhook config, outbound requests could be abused for SSRF (accessing internal network endpoints, metadata services, or localhost).

## Solution

Enforce URL validation and SSRF controls: allow only `https`, block private/link-local/loopback IP ranges after DNS resolution, optionally enforce a hostname allowlist, and disable redirects to disallowed targets.

## Changes

- `simpletuner/helpers/webhooks/config.py` (modified)

### DIFF
--- a/simpletuner/helpers/webhooks/config.py
+++ b/simpletuner/helpers/webhooks/config.py
@@ -1,6 +1,35 @@
+from ipaddress import ip_address
 from json import load
+from socket import IPPROTO_TCP, gaierror, getaddrinfo
+from urllib.parse import urlparse
 
 supported_webhooks = ["discord", "raw"]
+
+
+def validate_webhook_url(url: str, field_name: str):
+    parsed_url = urlparse(url)
+    if parsed_url.scheme != "https":
+        raise ValueError(f"Webhook config '{field_name}' must use https.")
+    if not parsed_url.hostname:
+        raise ValueError(f"Webhook config '{field_name}' must include a valid hostname.")
+    try:
+        resolved_addresses = {
+            address_info[4][0]
+            for address_info in getaddrinfo(parsed_url.hostname, parsed_url.port or 443, proto=IPPROTO_TCP)
+        }
+    except gaierror as e:
+        raise ValueError(f"Unable to resolve webhook hostname for '{field_name}': {e}")
+    for resolved_address in resolved_addresses:
+        parsed_ip = ip_address(resolved_address)
+        if (
+            parsed_ip.is_private
+            or parsed_ip.is_loopback
+            or parsed_ip.is_link_local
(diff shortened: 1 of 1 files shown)


bghira:
i can't really accept this change, the webhook infrastructure is used by the webui to receive events, often over non-ssl link as it's not required for localhost.

it's always been recommended that the webhook URLs be validated by the user at configuration time and that they do not publicly expose the service without adequate firewall protection.

can you demonstrate a real-world scenario in which this is exploitable, please?

tomaioo:
A realistic exploit is a multi-user/shared deployment (or a compromised low-priv account) where someone can edit webhook settings: they point `webhook_url` to `http://169.254.169.254/...` (cloud metadata) or `http://127.0.0.1:<admin-port>` and let normal events trigger server-side requests. Even if responses aren’t exposed, blind SSRF is still useful for internal port/service probing and can trigger state-changing internal endpoints. I agree localhost/non-SSL is needed for WebUI, so a practical compromise is “strict SSRF checks by default, with explicit localhost/allowlist exceptions for trusted admin configs.”

bghira:
fair, but there's probably some better guidance via documentation that's missing in the API tutorial doc (it has a dedicated one) and the ENTERPRISE.md doc which links to a lot of other stuff about multi-user environments and managing a production workspace.

i'm thinking some iptables rules that use uid tracking module would be best, eg. literally prevent the account from accessing anything except what it's supposed to. you can use AWS IAM roles to accomplish the same, probably?

so what will the change to the PR look like, we're going to exclude localhost and 127/8 from checks and add documentation?

tomaioo:
A realistic exploit is a multi-user/shared deployment (or a compromised low-priv account) where someone can edit webhook settings: they point `webhook_url` to `http://169.254.169.254/...` (cloud metadata) or `http://127.0.0.1:<admin-port>` and let normal events trigger server-side requests. Even if responses aren’t exposed, blind SSRF is still useful for internal port/service probing and can trigger state-changing internal endpoints. I agree localhost/non-SSL is needed for WebUI, so a practical compromise is “strict SSRF checks by default, with explicit localhost/allowlist exceptions for trusted admin configs.”

bghira:
i guess your harness is broken or something, it's repeating itself. i'm asking for a fixed pull request, or to abandon it. abandoning it is fine.

PR #2682, opened 2026-04-14
fix: replace bare except with explicit exception type in setup.py

harshadkhetpal:
## Summary

Replaces bare `except:` clauses with explicit exception types. Bare `except:` catches `SystemExit` and `KeyboardInterrupt`, which can mask test interruptions and swallow critical signals. This PR follows PEP 8 / E722 recommendations.

**Change:**
```python
# Before
try:
    ...
except:
    ...

# After
try:
    ...
except Exception:
    ...
```

For `except:` blocks that re-raise, `except BaseException:` is used instead so `SystemExit`/`KeyboardInterrupt` still propagate correctly.

## Testing

No behavior change for expected code paths. The change only affects how non-`Exception` signals (e.g. `SystemExit`, `KeyboardInterrupt`) propagate — making shutdown and Ctrl+C behave correctly.

## References

- [PEP 8 — bare except](https://peps.python.org/pep-0008/#programming-recommendations)
- [pycodestyle E722](https://www.flake8rules.com/rules/E722.html)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@ def run_command(cmd):
     try:
         result = subprocess.run(cmd, shell=True, capture_output=True, text=True)
         return result.returncode == 0
-    except:
+    except Exception:
         return False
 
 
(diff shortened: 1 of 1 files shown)


bghira:
sorry man, but i don't think you're a real user, and i don't think you've ever actually run simpletuner before. 

can you show a screenshot of your terminal for testing this change? thanks.

PR #2721, opened 2026-05-20
(#2569) add test to ensure CREPA loss alignment is only off for high-compression VAEs

### LINKED ISSUE #2569: CREPA: washed out results – bug or inherent limitation?
Hi, bghira!

I have been using CREPA for almost all my LTX-2 LoRAs and CREPA is a **massive** consistency / anatomy correctness booster. LTX-2 without any regulation is a disaster to train. Some LoRAs wouldn't train without it at all or they would retrain on particular concepts / visual elements, without getting the bigger picture. Thanks to CREPA I was able to train LoRAs in 1-5 hours on a 5090, not 9+ like people claim.

However, when I use it, whether it's a vanilla paper-compliant external VAE + DiNOv2 or the backbone features, the LoRA samples come slightly, or gradually washed out as the training goes on. The semantic consistency is a good tradeoff, but eventually the feel gets on.

There is an option to schedule CREPA, but maybe it's just a bug in the implementation? Papers like the original REPA, which everyone is using, or CREPA's paper have it as a constant (and a large one), yet they look good, like on the Disney's drawing dataset.

I noted the similarity log in the loss session transcends 1.0 and goes up, however, similarly should cap on 1.0, maybe it's a reason?

A some sort of similar washing out emergence happened in Kandinsky too, but I discarded it as insignificant.

bghira:
Closes #2569 

This isn't a bugfix for the problem of CREPA washing out samples, it's just a test to ensure that the problem is limited to high-compression video autoencoders.

The LTX-2 and Kandinsky5 Video VAEs have greater temporal compression than the CogVideo model used by the CREPA authors in their original paper. Trying to use the default tau and other values causes the semantic objective to cover too many frames.

No other documentation or tests or code have been updated.

### DIFF
--- a/tests/test_crepa.py
+++ b/tests/test_crepa.py
@@ -478,6 +478,74 @@ class CrepaFeatureSourceTests(unittest.TestCase):
             )
 
 
+class _IdentityProjector(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+        self._dtype_anchor = torch.nn.Parameter(torch.ones(1))
+
+    def forward(self, hidden_states):
+        return hidden_states
+
+
+class CrepaNeighbourSumTests(unittest.TestCase):
+    def _make_config(self, **kwargs):
+        defaults = {
+            "crepa_enabled": True,
+            "crepa_block_index": 0,
+            "crepa_adjacent_distance": 1,
+            "crepa_adjacent_tau": 1.0,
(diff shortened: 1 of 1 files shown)


MeiYi-dev:
Kandinsky 5 uses HunyuanVideo VAE which has the same compression ratio as the Wan2.1 VAE

PR #2848, opened 2026-07-17
trainingsample v0.3.0 for performance improvement on huge datasets

bghira:
This pull request updates the required version range for the `trainingsample` dependency from `>=0.2.18` to `>=0.3.0,<0.4.0` and adjusts the related test to match this new requirement.

Dependency updates:

* Updated the `trainingsample` dependency in `setup.py` to require versions `>=0.3.0` and `<0.4.0` instead of the previous `>=0.2.18`.

Test updates:

* Modified the test in `tests/test_setup_platform_dependencies.py` to check for the new `trainingsample` version range and ensure the old version requirement is not present.


bghira:
the improved tsr build brings total wall clock time on huge (2 million samples) datasets down from 4 hours to about 1 hour across 4x H200 and even better performance on gaming CPU class systems (eg. RTX 6000 Pro, RTX 5090, 4090) thanks to higher clock rates and boost freq

PR #2874, opened 2026-07-23
feat: add optional Kubeflow single-GPU worker integration

duroey:
## Summary

- Add an optional Kubeflow integration that creates one-GPU training Workers as Kubernetes TrainJobs.
- Keep the existing SimpleTuner Server and Worker protocol as the upper-layer abstraction; only Kubeflow jobs use the asynchronous Worker execution and final-artifact publishing path.
- Preserve the existing local scheduler, YAML dataloader handling, subprocess behavior, and S3 PUT response contract for non-Kubeflow jobs.
- Add targeted tests for Kubeflow job lifecycle, queue isolation, Worker credentials, artifact uploads, and compatibility regressions.

## Scope

This change intentionally supports one GPU per Kubeflow training job. It does not introduce private deployment manifests, internal registry references, or a replacement for SimpleTuner's control plane.

## Validation

Ran the targeted test suite in the existing SimpleTuner 4.5.0 Kubeflow image:

```text
179 passed
```

The suite covers the added Kubeflow services and routes plus the affected Worker, Trainer, publishing-provider, and S3 upload behavior.

### DIFF
--- a/documentation/OPTIONS.md
+++ b/documentation/OPTIONS.md
@@ -4,6 +4,21 @@
 
 This guide provides a user-friendly breakdown of the command-line options available in SimpleTuner's `train.py` script. These options offer a high degree of customization, allowing you to train your model to suit your specific requirements.
 
+### Kubeflow single-GPU server options
+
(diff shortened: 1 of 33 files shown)


bghira:
i don't have a kube flow cluster to test this on, have you actually verified that it works?

PR #2897, opened 2026-07-26
Fix distributed padding and preserve padded occurrences

hjinnkim:
## Summary

Fix distributed bucket partitioning for standard DP and Context Parallel paths while preserving intentional padded occurrences across sampler progress and checkpoint resume.

## Additional issues found

- The original standard DDP/FSDP example overstated the empty-tail behavior; that bug was specific to the custom CP splitter.
- Non-divisible CP splits did not match standard DP partition boundaries when padding was disabled.
- Path-based sampler progress collapsed intentional padded occurrences.
- Resume applied progress without first restoring the saved rank-local schedule.

## Resolution

- Use the same balanced quotient/remainder partition for standard DP and CP.
- Apply padding only to short rank-local shards after their real boundaries are established.
- Track repeated paths by occurrence count.
- Restore the saved rank-local schedule before applying resume progress.
- Preserve the same padding policy during epoch rollover.

Persistent discovery/cache deduplication remains unchanged. The separate automatic oversubscription scheduling issue is handled in #2918.

Regression coverage for the affected partition, padding, rollover, duplicate-occurrence, and resume cases is included.

### DIFF
--- a/simpletuner/helpers/data_backend/factory.py
+++ b/simpletuner/helpers/data_backend/factory.py
@@ -3283,7 +3283,7 @@ class FactoryRegistry:
                     f" You have to reduce your batch size, or increase your dataset size (id={init_backend['id']})."
                 )
 
-        apply_padding = True if not self.args.max_train_steps or self.args.max_train_steps == 0 else False
+        apply_padding = not self.args.max_train_steps or self.args.allow_dataset_oversubscription
 
         if backend.get("auto_generated", False):
             # when we're duplicating a metadata set, it's already split between processes.
--- a/simpletuner/helpers/metadata/backends/base.py
+++ b/simpletuner/helpers/metadata/backends/base.py
@@ -903,33 +903,21 @@ class MetadataBackend:
                         effective_batch_size=effective_batch_size,
                     )
 
-            # Split data by DP rank (not global rank) when context parallelism is enabled.
-            # This ensures all ranks in a CP group get the same data shard.
-            if cp_size > 1:
-                # Custom CP-aware splitting: split by dp_rank instead of process_index
-                chunk_size = ceil(len(trimmed_images) / effective_dp_size) if effective_dp_size > 0 else len(trimmed_images)
-                start_idx = dp_rank * chunk_size
-                end_idx = min(start_idx + chunk_size, len(trimmed_images))
-                images_split = trimmed_images[start_idx:end_idx]
-
-                # Handle padding if requested (for uniform batch sizes)
-                if apply_padding and len(images_split) < chunk_size and len(images_split) > 0:
-                    padding_needed = chunk_size - len(images_split)
-                    # Pad by repeating elements from the split
-                    padding = (images_split * ((padding_needed // len(images_split)) + 1))[:padding_needed]
-                    images_split = images_split + padding
-
-                new_aspect_ratio_bucket_indices[bucket] = images_split
-            else:
-                # Standard splitting when CP is disabled
-                with self.accelerator.split_between_processes(trimmed_images, apply_padding=apply_padding) as images_split:
-                    new_aspect_ratio_bucket_indices[bucket] = images_split
+            samples_per_rank, extra_samples = divmod(len(trimmed_images), effective_dp_size)
+            start_idx = dp_rank * samples_per_rank + min(dp_rank, extra_samples)
+            local_size = samples_per_rank + int(dp_rank < extra_samples)
+            images_split = trimmed_images[start_idx : start_idx + local_size]
+            if apply_padding:
+                target_size = samples_per_rank + int(extra_samples > 0)
+                if trimmed_images and len(images_split) < target_size:
+                    images_split += [trimmed_images[-1]] * (target_size - len(images_split))
+            new_aspect_ratio_bucket_indices[bucket] = images_split
 
         self.aspect_ratio_bucket_indices = new_aspect_ratio_bucket_indices
         post_total = sum([len(bucket) for bucket in self.aspect_ratio_bucket_indices.values()])
         if self.bucket_report:
             self.bucket_report.record_bucket_snapshot("post_split", self.aspect_ratio_bucket_indices)
-        if total_samples != post_total:
+        if self.accelerator.num_processes > 1 or total_samples != post_total:
             self.read_only = True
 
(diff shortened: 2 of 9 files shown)


bghira:
@hjinnkim to be honest i'm not sure this change is going to work as you intended, we deduplicate the lists and so padding won't actually work the way your PR says it would

bghira:
@copilot can you investigate the complications of how we're deduplicating aspect buckets especially where we're converting lists into sets to ensure we collapse duplicate entries. there was a problem with apply_padding before which caused this stack trace:

```
Traceback (most recent call last):
  File "/workspace/SimpleTuner/train.py", line 50, in <module>
    trainer.init_data_backend()
  File "/workspace/SimpleTuner/helpers/training/trainer.py", line 678, in init_data_backend
    raise e
  File "/workspace/SimpleTuner/helpers/training/trainer.py", line 652, in init_data_backend
    configure_multi_databackend(
  File "/workspace/SimpleTuner/helpers/data_backend/factory.py", line 1229, in configure_multi_databackend
    init_backend["vaecache"].build_vae_cache_filename_map(
  File "/workspace/SimpleTuner/helpers/caching/vae.py", line 181, in build_vae_cache_filename_map
    for image_file in all_image_files:
TypeError: 'NoneType' object is not iterable
```

in April 2025. the user who reported that issue in that case said the problem manifests on a fresh system with no pre-existing file cache. the cache prevents the symptom, which makes it look like the problem isn't there. but i'm curious what kind of "race condition" could cause this - and, no, dist barriers didn't help

bghira:
from Codex:

```
  The “race condition” was likely not a distributed training race. It looks like a file-list cache publication race:

  - StateTracker._save_to_disk() opens JSON with "w" before taking the lock, which truncates the file before the exclusive lock is held: simpletuner/helpers/training/state_tracker.py:155
  - _load_from_disk() returns None immediately on JSON/load errors instead of retrying: simpletuner/helpers/training/state_tracker.py:130
  - Older code could then pass that None into build_vae_cache_filename_map(), which blindly iterates it: simpletuner/helpers/caching/vae.py:689

  That explains “fresh system only”: no pre-existing all_image_files_*.json means subprocesses are dependent on first-run cache creation. A barrier does not make a non-atomic filesystem write visible atomically, and it does not help if a process already cached/observed None.

  On current main, the exact stack trace is probably already mitigated by the fallback in the VAE setup path: if StateTracker.get_image_files(..., retry_limit=3) returns None, factory lists files directly and calls StateTracker.set_image_files() before building the map: simpletuner/
  helpers/data_backend/factory.py:4186.

  For PR 2897 specifically, the CP padding change is directionally correct. The old CP branch only padded non-empty local shards, so an undersized bucket could still leave a DP shard empty: simpletuner/helpers/metadata/backends/base.py:904. Repeating the final global item matches
  Accelerate’s apply_padding=True behavior better.

  The dedupe complication is separate: padded duplicate bucket entries are intentional, while discovery/cache duplicates are not. These should not be collapsed by the same logic. In particular, update_buckets_with_existing_files() removes duplicates with dict.fromkeys(...): simpletuner/
  helpers/metadata/backends/base.py:961. That is fine for persistent discovered buckets, but it would undo intentional post-split padding if applied after splitting. The VAE cache can dedupe for storage because one latent per source file is enough, but training bucket shard length must
  preserve intentional duplicates.

  I’d ask PR 2897 to add tests. The PR body says verified, but the diff contains no tests. At minimum:

  - non-CP oversubscribed fixed-step bucket smaller than effective DP size keeps every rank non-empty
  - CP oversubscribed bucket pads empty DP shards from the final global item
  - a bucket refresh/cache cleanup does not collapse intentional post-split padding duplicates
```

bghira:
I inspected this because it touches distributed bucket splitting / resume-fragile data behavior. I think there is a real bug here, but the PR needs tightening before merge.

What I verified:

- `FactoryRegistry._handle_bucket_operations()` currently ignores `allow_dataset_oversubscription` when `max_train_steps` is set, so explicit oversubscription can auto-adjust repeats but still split without padding.
- Passing the same padding policy through epoch rollover makes sense; rebuild/re-split should not use a different policy than initial dataset construction.
- The CP branch in `MetadataBackend.split_buckets_between_processes()` has a real empty-tail-shard bug when `apply_padding=True`: current code only pads non-empty `images_split`, so tail DP ranks can remain empty. Repeating `trimmed_images[-1]` fixes that specific CP case.

Concerns before merge:

- The PR body's 4-images / 8-GPU / no-CP example is misleading. For the normal DDP/FSDP path, Accelerate's `split_between_processes(..., apply_padding=False)` already returns the final input item for ranks whose start index is past the end. The empty-tail-shard behavior is in SimpleTuner's custom CP splitter, not the standard splitter. Please correct the description so reviewers are not validating the wrong behavior.
- The branch does not include regression tests despite the body saying focused cases passed. Please add tests for the factory padding predicate, epoch rollover forwarding the same policy, and CP empty-tail shards under `apply_padding=True` / `allow_dataset_oversubscription=True`.
- There is still a CP no-padding parity edge case outside this PR's narrow fix: with 9 samples, effective DP size 8, and `apply_padding=False`, Accelerate/DDP gives local lengths `[2,1,1,1,1,1,1,1]`, while the custom CP splitter gives `[2,2,2,2,1,0,0,0]`. If this PR intentionally only fixes explicit oversubscription, that scope should be called out clearly.

I would hold merge until the tests are added and the PR body is corrected. The underlying oversubscription + CP padding fix looks plausible, but the current justification overstates the standard DDP/FSDP behavior.

hjinnkim:
Updated #2897 after reviewing the deduplication and distributed-split concerns.

Additional issues found:
- non-divisible CP splits differed from standard DP behavior;
- path-based sampler progress collapsed intentional padded occurrences;
- resume needed the saved rank-local schedule restored before applying progress.

The update now uses one balanced partitioning policy for standard DP and CP, preserves repeated occurrences across progress/resume, and leaves persistent discovery/cache deduplication unchanged.

The separate automatic oversubscription scheduling issue is split into #2918.

hjinnkim:
Sorry for force-pushing this branch while the PR was under active review. The branch history was updated from `dc1cb91` to `3637aa1`, which may have affected some existing review context.

I’m currently working on resolving the remaining issues in this PR. The force-push happened accidentally during that process.

PR #2898, opened 2026-07-26
Fix deterministic distributed bucket splitting

hjinnkim:
# Fix deterministic distributed bucket splitting

## Summary

- Process-local RNG state could give ranks different permutations of the same bucket before splitting.
- When each rank slices a different permutation, distributed training can duplicate some samples and omit others in an epoch.
- Use a shared shuffle seed and a dataset/bucket-specific RNG so every rank slices the same bucket order.

## Changes

- `simpletuner/helpers/metadata/backends/base.py`: reuse the configured seed, or generate one on the main process and broadcast it; shuffle a bucket copy with a dataset-and-bucket-specific RNG before rank slicing.
- `simpletuner/helpers/metadata/backends/base.py`: avoid a second full-list copy when the trim limit keeps the complete bucket.
- `AGENTS.md`: document that checkpoint resume is not assumed to support topology or dataloader-setting changes between save and resume.

## Notes

- This intentionally does not restore saved sampler partitions or add compatibility handling for changed DDP/FSDP/CP topology, world size, batch sizing, gradient accumulation, repeats, bucket splitting, shuffling, or dataloader configuration. Those changes may produce unexpected behavior when resuming and need explicit support before being treated as safe.

## Verification

```bash
/Users/kash/src/SimpleTuner/.venv/bin/python -m unittest tests.test_factory_edge_cases.TestFactoryEdgeCases.test_early_validation_sufficient_repeats tests.test_factory_edge_cases.TestFactoryEdgeCases.test_eval_dataset_ignores_gradient_accumulation tests.test_factory_edge_cases.TestFactoryEdgeCases.test_oversubscription_auto_adjustment -v
```

Also verified with a focused 8-rank bucket split probe: 1,024 assigned, 1,024 unique, 0 missing, 0 duplicated entries.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -40,6 +40,7 @@
 - Problems should always be provable through tests, logging, or other means
 - Generally speaking, it's fine to run the full application end-to-end to verify a fix, "it's heavy" is not a valid excuse to avoid verification - we're on a ML development workstation that's designed to allow running these workloads
 - For the most part, things should not be marked as CUDA-only unless it relies on third-party compiled CUDA kernels or similar. Don't be afraid to use the available accelerator eg. mps, cuda, if available on the system opportunistically.
+- Do not assume checkpoint resume supports changing topology or dataset settings. Changing DDP/FSDP/CP layout, world size, batch sizing, gradient accumulation, repeats, bucket splitting, shuffling, or dataloader configuration between save and resume can produce unexpected behavior; avoid adding compatibility fallbacks unless support for that specific change is explicitly required and verified.
 
 ## Frontend testing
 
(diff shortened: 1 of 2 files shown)


bghira:
i removed the sampler restore changes from your pull request because they're unexpectedly complex and fragile. i added guidance to AGENTS.md so that future sessions are unlikely to go down the path of resolving bugs that centre around unsupported use like altering training hardware layout or config.json (in many cases, at least). this is because changing these things can often cause unexpected behaviour anyway. it's difficult to verify at every level of the stack that adjusting something mid-training won't cause a problem.

PR #2899, opened 2026-07-26
Fix per-rank sampler state resume path

hjinnkim:
# Fix per-rank sampler state resume path

## Summary

- Resume rebuilt the per-rank sampler state filename independently from the save hook.
- Nonzero ranks saved through `SaveHookManager.training_state_path`, e.g. `training_state-rank1.json`, but sampler resume looked for `training_state-1.json` before backend state path mangling.
- Load sampler state from the hook-owned `self.model_hooks.training_state_path` instead.

## Changes

- `simpletuner/helpers/training/trainer.py`: use `SaveHookManager.training_state_path` when loading sampler state on checkpoint resume.
- `tests/test_trainer.py`: add a regression test for nonzero-rank sampler resume loading `training_state-rank1.json`.

## Notes

- This intentionally leaves completed-partition sampler rollover behavior unchanged. A fully consumed sampler state can be the epoch-end signal used to coordinate ranks before the next training step.
- Model, optimizer, and scheduler resume behavior is unchanged.

## Verification

```bash
/Users/kash/src/SimpleTuner/.venv/bin/python -m unittest tests.test_trainer.TestTrainer.test_init_resume_checkpoint tests.test_trainer.TestTrainer.test_init_resume_checkpoint_loads_sampler_from_hook_training_state_path -v
```

### DIFF
--- a/simpletuner/helpers/training/trainer.py
+++ b/simpletuner/helpers/training/trainer.py
@@ -4687,16 +4687,13 @@ class Trainer:
             job_id=self.job_id,
         )
         self._emit_event(event)
-        training_state_filename = f"training_state.json"
-        if get_rank() > 0:
-            training_state_filename = f"training_state-{get_rank()}.json"
         for _, backend in StateTracker.get_data_backends().items():
             if "sampler" in backend:
                 backend["sampler"].load_states(
                     state_path=os.path.join(
                         self.config.output_dir,
                         path,
-                        training_state_filename,
+                        self.model_hooks.training_state_path,
                     ),
                 )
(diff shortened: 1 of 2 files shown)


bghira:
i split the filename fix from the sampler change and removed the sampler part. i'd like for that to be submitted separately with some demonstration that it doesn't cause a barrier wait condition. the thing is, the MultiDatasetExhausted signal is really the final say on the sampler resetting its buckets, and when running ddp, an all-reduce is done across nodes/GPUs to determine if any rank had an exhaustion signal. if they do, then they all roll to the next epoch so that things are kept in sync. can you please explore that for the next time the sampler change is submitted? thank you for your pull requests and improvements you've provided so far.

PR #2918, opened 2026-07-27
Fix automatic oversubscription scheduling with gradient accumulation

hjinnkim:
## Dependency

Depends on #2897 for occurrence-aware sampler and resume semantics. This branch is stacked on #2897 and will be rebased after it merges.

## Additional issues found

- Automatic oversubscription increased the repeat count without constructing the repeated schedule, so gradient accumulation could still leave insufficient rank-local samples.
- `repeats: 0` was treated as a manual override.
- Empty buckets could enter repeat division.
- Mutating the backend repeat value could affect later splits.
- Conditioning metadata copying could discard the current rank-local source schedule and overwrite canonical target bucket state with rank-local data.

## Resolution

- Keep the automatic repeat calculation local to the split.
- Build a batch-aligned cyclic schedule directly at rank-local size.
- Treat only positive configured repeats as a manual override.
- Skip empty buckets.
- Preserve the source schedule and avoid persisting rank-local target bucket indices over canonical metadata.

Regression coverage for gradient accumulation, standard DP, Context Parallel, empty buckets, manual repeats, and conditioning metadata is included.

### DIFF
--- a/simpletuner/helpers/data_backend/factory.py
+++ b/simpletuner/helpers/data_backend/factory.py
@@ -3283,7 +3283,7 @@ class FactoryRegistry:
                     f" You have to reduce your batch size, or increase your dataset size (id={init_backend['id']})."
                 )
 
-        apply_padding = True if not self.args.max_train_steps or self.args.max_train_steps == 0 else False
+        apply_padding = not self.args.max_train_steps or self.args.allow_dataset_oversubscription
 
         if backend.get("auto_generated", False):
             # when we're duplicating a metadata set, it's already split between processes.
--- a/simpletuner/helpers/metadata/backends/base.py
+++ b/simpletuner/helpers/metadata/backends/base.py
@@ -778,10 +778,17 @@ class MetadataBackend:
         if self.bucket_report:
             self.bucket_report.set_constraints(effective_batch_size=effective_batch_size)
 
+        backend_config = StateTracker.get_data_backend_config(self.id) or {}
+        configured_repeats = int(backend_config.get("repeats") or 0)
+        user_set_repeats = configured_repeats > 0
+        auto_repeat_count = None
+
         # Early validation: check if configuration is mathematically impossible
         buckets_that_will_fail = []
         for bucket, images in self.aspect_ratio_bucket_indices.items():
             total_img_count_incl_repeats = len(images) * (self.repeats + 1)
+            if not images:
+                continue
             if total_img_count_incl_repeats < effective_batch_size:
                 buckets_that_will_fail.append(
                     {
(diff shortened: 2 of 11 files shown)


bghira:
@hjinnkim ready?

hjinnkim:
While validating this branch against the merged #2897 codebase I hit two issues on current `main` (`321457a8`) that this PR already resolves. Adding the details here so the coverage in the new commit is easy to follow.

**1. An empty aspect bucket key crashes startup**

`update_buckets_with_existing_files` assigns `[]` to a bucket whose files all disappeared rather than dropping the key, and the bare `save_cache()` that follows uses `enforce_constraints=False`, so the empty key is written to disk and reloaded on the next run. The next split then runs `ceil(effective_batch_size / images)` with `images == 0`.

`--allow_dataset_oversubscription` cannot avoid it, because the division happens before the auto-adjust branch that flag controls. It is not multi-GPU specific either: `effective_batch_size` is at least 1, so an empty bucket always joins `buckets_that_will_fail`.

> Repro on `321457a8`: a backend holding `{"1.0": ["a.jpg", "b.jpg"], "1.5": []}` raises `ZeroDivisionError: division by zero` out of `split_buckets_between_processes` for world sizes 1, 2 and 8, with `--allow_dataset_oversubscription` both on and off.

**2. Duplicating metadata discards the source's rank-local schedule**

A multi-GPU split marks the backend read-only, so its post-split `save_cache` is a no-op and the cache on disk stays the unsplit discovery list. `DatasetDuplicator.copy_metadata` then reloaded that list back over the source's shard; the restore guard there only fires when the reload comes back empty, which cannot happen once a cache file exists. The conditioning dataset inherits the clobbered list, and because the target is marked read-only the factory skips its own split, so nothing re-shards it. This is reached by any auto-generated or `reference_strict`/`mask` conditioning dataset — ControlNet included — on more than one GPU.

> Repro on `321457a8`: with 4 images across 8 ranks and padding enabled, every rank ends up holding the full 4-image list after `copy_metadata` instead of its 1-image shard (`[4] * 8` where `[1] * 8` is expected).

**Both are fixed by merging this PR as it stands.** The empty-bucket guard in `split_buckets_between_processes` and the `copy_metadata` rework already on this branch — dropping the source reload, and marking the target read-only before anything is persisted — cover them; no further production change is needed.

**Regression coverage**

I pushed a test-only commit (`943d8373`, no production changes) that pins exactly these two behaviours:

- `tests/test_metadata_backend.py::TestEmptyBucketRepeatValidation` — the empty-bucket split across world sizes 1/2/8 and both oversubscription settings, with the pre-existing "zero usable batches" error preserved for the flag-off case, plus the `update_buckets_with_existing_files` behaviour that leaves the empty key behind in the first place.
- `tests/test_conditioning_split_alignment.py::TestConditioningSplitAlignment::test_duplication_preserves_the_padded_split_source_schedule` — all eight ranks run a real padded split with the unsplit source cache present on disk, then duplicate; it asserts every rank keeps its shard, that the shards still cover the dataset, that the conditioning dataset inherits the same shard, and that the shared conditioning cache never becomes rank-dependent.

Run against current `main` these fail with the `ZeroDivisionError` and the `[4] * 8` schedule above; on this branch they pass. Two notes so the results are not surprising: `test_refresh_leaves_an_empty_bucket_key_behind` passes on both, since it pins the producer rather than the crash, and `test_bucket_without_empty_keys_is_unaffected` fails on `main` for an unrelated reason — it asserts this PR's semantics, that automatic oversubscription leaves `repeats` unmutated.

Since #2897 is merged, this branch is based on a commit that is already in `main`, so it needs no rebase and merges cleanly as-is.

PR #2923, opened 2026-08-02
Update example presets

bghira:
Summary:
- add ACE Step audio duration bucketing defaults to the example backend
- correct model_family and LyCORIS config path values for several examples
- update example precision, attention, checkpointing, and optimizer settings where the presets were stale
- remove invalid trailing commas from affected JSON examples

Validation:
- jq empty on all changed JSON files
- rg scan for /Users paths and LLM/session markers
- git diff --cached --check
- pre-commit hooks from git commit


bghira:
Superseded by model-specific example preset PRs: #2964 through #2975.

PR #3001, opened 2026-08-05
Extend metadata-only bucketing to still images in the discovery backend

AL3708:
### Context

Metadata-only bucketing already exists in SimpleTuner — but not for still
images scanned from a local directory.

Where it does exist:

- `discovery.py` skips decoding **videos** via ffprobe
  (`_should_use_metadata_only_for_video` → `use_metadata_only = True`)
- the `parquet` backend reads `width_column` / `height_column`
- the `huggingface` backend reads width/height dataset columns
- the `webshart` backend reads dimensions from shard metadata

The gap: with the `discovery` backend on a plain local image folder — the most
common setup — every image is fully read and decoded just to reach
`image.size`:

```python
image_data = self.data_backend.read(image_path_str)
image = load_image(BytesIO(image_data))
...
image_metadata["original_size"] = image.size
```

Bucketing never needs those pixels. `calculate_target_size()` and the
crop-coordinate maths in `TrainingSample` derive everything from
`original_size`, and `PreparedSample` stores no pixel-derived metadata. On a
network filesystem this turns discovery into a multi-megabyte read per image
for data that is immediately discarded — and the only way to avoid it today is
to build and maintain an external parquet/JSONL of dimensions the file headers
already contain.

### Change

Still images now take the same metadata-only path as videos, reading
dimensions from the file header.

`use_metadata_only` already supports this downstream — `TrainingSample`
tolerates `image=None`:

- `crop()` guards its pixel work on
  `self.image is not None and isinstance(self.image, Image.Image)`
- the RGB/EXIF transform helper guards on `self.image is not None`
- `meets_resolution_requirements()` has an `image_metadata` branch reading
  `original_size`

So this only adds the entry point, plus a header probe.

The two per-media guards are folded into one
`_should_use_metadata_only(is_video_file)`, so the face-crop exclusion lives in
a single place rather than being duplicated. Videos keep the ffprobe route;
images take the header route. `is_video_file` is passed explicitly so a
video-extension file inside an `IMAGE` dataset still falls back to a full
decode, exactly as before.

`_probe_image_dimensions()` opens the file with PIL, which parses only the
header — pixel data is loaded lazily and never requested.

Guards, following the existing video precedent:

- local data backends only (a header read needs a real path)
- `IMAGE` and `CONDITIONING` dataset types
- **not** when crop is enabled with `crop_style="face"` — the one crop style
  whose coordinates depend on pixel content
- any header read failure returns `None` and falls back to a full decode, so
  truncated, corrupt or unsupported files behave as before

### EXIF correctness

EXIF orientations 5–8 transpose the image, so header dimensions are swapped
relative to what `exif_transpose()` yields on the decode path. The orientation
tag is in the header too, so it is applied in the probe.

Verified equal to the full-decode result across:

| Case | probe | full decode |
|---|---|---|
| no EXIF | (1200, 800) | (1200, 800) |
| orientations 1–4 | (1200, 800) | (1200, 800) |
| orientations 5–8 | (800, 1200) | (800, 1200) |
| PNG / WebP / 1×1 JPEG | match | match |
| corrupt header | `None` → fallback | — |

### DIFF
--- a/simpletuner/helpers/metadata/backends/discovery.py
+++ b/simpletuner/helpers/metadata/backends/discovery.py
@@ -6,6 +6,8 @@ import traceback
 from io import BytesIO
 from typing import Optional
 
+from PIL import Image
+
 from simpletuner.helpers.data_backend.base import BaseDataBackend
 from simpletuner.helpers.data_backend.dataset_types import DatasetType
 from simpletuner.helpers.image_manipulation.load import load_image, load_video
@@ -94,16 +96,58 @@ class DiscoveryMetadataBackend(MetadataBackend):
             max_num_samples=max_num_samples,
         )
 
-    def _should_use_metadata_only_for_video(self) -> bool:
-        if not _is_ffprobe_available():
-            return False
-        if self.dataset_type is not DatasetType.VIDEO:
-            return False
+    def _should_use_metadata_only(self, is_video_file: bool) -> bool:
+        """Whether this sample can be bucketed from dimensions alone.
+
+        Bucketing needs only the sample dimensions: calculate_target_size() and
+        the crop coordinate maths derive everything from original_size, and no
+        pixel-derived metadata is stored. Videos already take this path via
+        ffprobe; still images can do the same from the file header.
+
+        Face cropping is the one crop style whose coordinates depend on pixel
+        content, so it always requires a full decode.
+        """
         crop_enabled = bool(self.dataset_config.get("crop", False))
         crop_style = str(self.dataset_config.get("crop_style") or "random").lower()
         if crop_enabled and crop_style == "face":
             return False
-        return True
+
+        if is_video_file:
+            return self.dataset_type is DatasetType.VIDEO and _is_ffprobe_available()
+        if self.dataset_type in (DatasetType.IMAGE, DatasetType.CONDITIONING):
+            # Header reads need a real path; remote backends stay on full reads.
+            return getattr(self.data_backend, "type", None) == "local"
+        return False
+
+    def _probe_image_dimensions(self, image_path_str: str) -> Optional[dict]:
+        """Read image dimensions from the file header, without decoding pixels.
+
+        PIL parses only the header on Image.open(); pixel data is loaded lazily
+        and never requested here. Returns None to signal that the caller should
+        fall back to a full decode.
+
+        EXIF orientations 5-8 transpose the image, so the stored dimensions are
+        swapped relative to what exif_transpose() produces on the full-decode
+        path. The orientation tag lives in the header too, so it is applied here
+        to keep both paths in agreement.
+        """
+        try:
+            with Image.open(image_path_str) as image:
+                width, height = image.size
+                orientation = image.getexif().get(0x0112)
+            if orientation in (5, 6, 7, 8):
+                width, height = height, width
+        except Exception as e:
+            logger.debug(
+                "(id=%s) Could not read image header for %s (%s); falling back to full decode.",
+                self.id,
+                image_path_str,
+                e,
+            )
+            return None
+        if not width or not height:
+            return None
+        return {"original_size": (width, height)}
 
     def _needs_video_frame_count(self) -> bool:
         if self.bucket_strategy == "resolution_frames":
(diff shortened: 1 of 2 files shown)


bghira:
unfortunately, PIL is really slow for decode versus TrainingSample which uses OpenCV under the hood. can you place this path through tsr instead? might need to open a PR to the TrainingSample repo to add a probe helper.

AL3708:
Hi,
Avoiding any decode at all is the goal here — bucketing doesn't need the pixels, only `original_size`.

`Image.open()` is lazy: it parses the container header to populate `size`/`mode`/`format` and returns. Pixel data is only read on `.load()`, `.convert()`, or array conversion. `_probe_image_dimensions()` touches `.size` and EXIF tag 0x0112 — both header fields — and never triggers a load. On a 4 MB JPEG that's a ~200 byte read, against reading and decompressing the whole file on the path it replaces. No decoder runs, so decode speed doesn't enter into it.

Where a decode genuinely is needed, nothing changes: the probe returns `None` on any header read failure or unsupported format, and the existing full-decode path handles it exactly as before. Same for face cropping and non-local backends, which are excluded from the metadata-only path.

That makes this one-directional — it can only speed things up, never slow them down. On the hit path a full read+decode becomes a header read. On the fallback path the only added cost is that header read itself before the original code runs, which is negligible next to the decode that follows. There's no configuration where this does more work than before.

bghira:
well, one thing that the full decode would achieve is to allow us to delete corrupted samples if eg. `--delete_problematic_images` is provided and enabled. i guess this is caught later on during `read_image`, and in VAE Cache as `read_image_batch` and the delete call is executed by each databackend. however, i'd request that this be checked before merge. the change here would still allow deletion if we fall through to the full read path, but in case it doesn't fall through (eg. headers show size but the file is truncated and not able to be decoded) i'd like to know it won't cause further problems.

AL3708:
Spent some time on testing this. You were right to push back, and it also turned up two things that already affect main.

Why I care about not just reverting to a full decode: I have ~700k images on NFS. Full-decode discovery is an overnight job, header-only is about 15 minutes.

**The bug is real**, though narrower than "truncation" — it's files where the header survives and the pixels don't. Probe succeeds, fallback never fires, decode blows up later. I truncated a test image in every format PIL writes: png, uncompressed tif, bmp, gif, jp2, tga, ppm and qoi all report correct dimensions and then fail to decode. webp, LZW tif and avif fail at the header, so those fall back fine.

**Two things already on main:**

- jpg looks clean only because `tsr.imdecode_py` doesn't care — it returned a full-size array for a JPEG cut to 5% of its bytes. So even with a full decode, truncated JPEGs sail through discovery today and `--delete_problematic_images` never fires. They just train as garbage.
- Video has the same header-vs-pixels gap. `moov` position decides it: a plain truncated mp4 fails at the header, but with `-movflags +faststart` (standard for web delivery) ffprobe returns complete metadata for a file that won't decode, and it gets bucketed.

**Three ways to fix it, and I don't want to guess wrong:**

1. Gate the image path on `--delete_problematic_images`. One-liner, but useless without the flag set, and that's the case where discovery used to quietly skip the file.
2. Make the fast path opt-in (per-dataset `metadata_only_scan`, default off). Safe, but nobody gets the speedup unless they go looking.
3. Fix the vae.py reconciliation so the batch path drops what it couldn't read. Root cause, safe by default, and fixes the `+faststart` video case too — but it's shared caching code this PR doesn't otherwise touch.

What do you think?

bghira:
vae cache is a pretty safe bet but some pixel models wouldnt be hitting that path at all - and vae cache can be ondemand or disabled, which means..

hmm.. maybe batch fetcher and prefetch especially can be modified to delete problematic files (if enabled), log warning, and skip that batch element, only raising a runtime error if errored file count being dropped means now this training run has no data or less than a single batch worth (as long as dataset oversubscription isnnt enabled)

thanks for checking that, it sounds like a mild bug in tsr as well?

bghira:
the tsr bug is fixed in [this PR](https://github.com/bghira/TrainingSample/pull/38) to reject broken jpg files

PR #3093, opened 2026-08-16
Guard batch-size changes when resuming sampler state

bghira:
## What
- Compare the checkpointed sampler batch size with the current per-dataset batch size before restoring state.
- Fail with an actionable error on mismatch unless the user explicitly enables `--i_know_what_i_am_doing` or `SIMPLETUNER_ALLOW_MODIFYING_BSZ=1`.
- Log a warning when that override permits the mismatched resume.
- Document the environment override and flag behavior in OPTIONS and all translations.
- Preserve legacy checkpoint behavior when no batch size was recorded.

## Why
Sampler checkpoints persist rank-local bucket schedules, exhausted buckets, cursors, and seen-image state generated for a specific batch size. Restoring that state under a different batch size is risky and should require an explicit opt-in.

## Impact
Resumes with matching settings are unchanged. A changed global default or dataset override fails before any live sampler state is mutated unless the explicit unsafe-resume override is enabled.

## Validation
- `.venv/bin/python -m unittest -f tests.test_sampler` (25 passed)
- `git diff --check`


bghira:
@copilot let's allow this if --i_know_what_im_doing is provided though or a specific env var, SIMPLETUNER_ALLOW_MODIFYING_BSZ and document that in OPTIONS.md and translations

PR #3130, opened 2026-08-21
Add DiffusionBlocks training support

### LINKED ISSUE #3110: [Feature Request] DiffusionBlocks: Block-wise Neural Network Training via Diffusion Interpretation (Significant reduction in VRAM / RAM usage while also improving the end result)
https://arxiv.org/abs/2506.14202

https://github.com/SakanaAI/DiffusionBlocks

https://sakana.ai/diffusion-blocks/

This would allow 12GB-16GB VRAM and low RAM users (~65% of the users) to train significantly large (Flux 2 Dev 32B, Cosmos3 32B, Minimax H3 32B, etc) DiT models.

Image

A case could also be made to just train a whole block for an epoch at a time so constant reads don't happen, allowing for much larger batch sizes (meaning much faster training) and allowing us to even maybe deprecate other block swapping methods.

This also opens up the DiT arch to be trained in a decentralized way where each PC is training a singular/a set of blocks.

An optimal implementation would be with the following settings:

Number of layers per block - Sets the amount of layers per block.

Blocks to train - Sets the block/s to train (This allows for decentralized runs/multi day runs where the PC rests at night, exporting a LoRA/LoKR that modifies the block's layers only)

TREAD, CREPA with ViT should also just work.

Image

Image

bghira:
## Summary
- add generic DiffusionBlocks routing for homogeneous diffusion Transformer stages with flow and discrete timestep partitioning
- support independent block ownership, DDP unused-parameter handling, physical layer indices, and clipped global TREAD routes
- reject incompatible architectures and fixed-layer auxiliary losses during setup
- add WebUI configuration, an Anima v1.0 example, and complete translated documentation

## Verification
- `.venv/bin/python -m unittest -v -f` (6,220 tests passed; 196 skipped)
- `.venv/bin/mkdocs build --strict`
- Anima v1.0 one-step H200 training smoke at 1024 resolution; seven routed groups initialized, schedule boundaries resolved, checkpoint saved, peak GPU memory 5.81 GiB

Closes #3110


bghira:
@copilot+gpt-5.3-codex can you like plz review kthx

